PR #239, opened 2016-04-01
Also restore task description when applying a template

eocanha:
Please, merge this commit if you think it's interesting. Delete the branch if not.


jaragunde:
The `text` field in template actually stores the template name (to show it in the button), so restoring it to the task description will not be very useful.

Using that field as the template name was a quick hack to avoid the need to add any specific UI for templates. The proper way to fix this would be to add the `title` field to the template and use `text` for its original purpose, the task description.

We could open a JS popup when the **template** button is pressed to ask users for a title for the template.

jaragunde:
This part of the code is being reworked in #254, so the patch is no longer vaild.

Issue #133, opened 2016-03-09
Open projects from the projects evaluation list in a new window


jaragunde:
[Trac import]
Reported by: xcastanho
Original date: Wednesday, 12 January 2011 13:04

When you have configured some filtering in the project evaluation page, there is a problem when double clicking on a project, because if you go back to the evaluation page the filter options would disappear.

Just giving some way to open the project details page in a new browser window/tab would solve the problem.

jaragunde:
[Trac import]
Comment by: jaragunde
Original date: Wednesday, 26 November 2014 18:38

```
commit 1c7f38f597230b65566a084214b95c5e6e004724
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Wed Nov 26 18:28:37 2014 +0100

    [#133] Always open project details in a new window.
    
    Double-clicking on a project from a report is often done to check some
    details of a project, but that doesn't mean you want to close the
    report on screen.
    
    This solution is simple and prevents losing time re-entering report
    filtering data again, at the cost of having to close a tab.
```

PR #257, opened 2016-07-12
[#238] Add button to go to project details from project evaluation report

### LINKED ISSUE #238: Add button to go to project details from project evaluation report
[Trac import]
Reported by: jaragunde
Original date: Wednesday, 16 December 2015 17:04

Right now you can double click on a project to go to the project details report from the project evaluation report, but there should be a button to make this feature discoverable.

tonythomas01:
Functionality copied from viewProjects.php. Fix #238


jaragunde:
Hi! The patch works but I think it's a bit too complex, probably because the code in `viewProjects.php` was already too complex. I'll leave you some comments inline.

web/projectsEvaluation.php
```diff
@@ -394,13 +394,41 @@ function loadProjects() {
             text: 'Extended view',
             handler: showExtendedView,
         }],
+        initComponent: function() {
+            this.tbar = this.buildTopToolbar();
+            Ext.ux.ExportableGridPanel.superclass.initComponent.apply(this, arguments);
+        },
```
jaragunde: This method is only required to call `buildTopToolbar()` inside the context of the `projectGrid` object, so `this` refers to it. Let's do things more simple by removing the entire `initComponent` method and doing things a bit differently.

web/projectsEvaluation.php
```diff
@@ -394,13 +394,41 @@ function loadProjects() {
...
+            return [{
+                text: 'Details',
+                id: this.id + 'DetailsBtn',
+                ref: '../detailsBtn',
+                disabled: true,
+                iconCls: this.iconCls + '-go',
+                handler: this.onDetails,
+                scope: this
+            }]
+        },
```
jaragunde: Instead of returning this array and assigning it to the `tbar` property in line 398, do that assigment here instead:
```
tbar: [{
    text: 'Details',
    ...
}],
```
You will have to remove any references to `this`:
* iconCls: use the string it refers to.
* handler: see the comment below.
* scope: I think you don't need it.

web/projectsEvaluation.php
```diff
@@ -394,13 +394,41 @@ function loadProjects() {
...
+        /**
+         * onDetails
+         */
+        onDetails: function() {
+            if (this.getSelectionModel().getCount() > 0) {
+                var selected = this.getSelectionModel().getSelected();
+                window.location = 'viewProjectDetails.php?pid=' + selected.id;
+            }
+        }
```
jaragunde: Take this `onDetails` function and register it outside the `projectGrid` object, like other button handlers like `showStandardView`. Replace any references to `this` inside the function by references to `projectGrid`.

jaragunde: And please, open the project details in a new window. That was an explicit user request in #133.

tonythomas01:
Fixed now. Looks good ?

Issue #225, opened 2016-03-09
Tasks created too soon cannot be saved


jaragunde:
[Trac import]
Reported by: jaragunde
Original date: Monday, 23 December 2013 11:53

If you press the new task or a template button to create a task before the load operation has been completed, that task is not properly added to the list and it cannot be saved.

Steps to reproduce:
1. Go to some date that contains saved tasks.
2. Place the mouse over one of the new task buttons.
3. Reload and press the button immediately, before the tasks are loaded.
  * If you did it correctly, the new task will be the first one in the list.
4. Fill the new task and press save. Nothing happens.

jaragunde:
To easily reproduce this bug, add a [sleep()](http://php.net/manual/en/function.sleep.php) call in the `getUserTasks` service. You will have all the extra time you want to create the new task following the steps above.

tonythomas01:
I find this from the apache error logs on trying to save. Might be uesful. 
`[Tue Jul 12 22:43:10.620874 2016] [:error] [pid 17792] [client ::1:47641] PHP Notice:  Undefined index: sid in /var/www/igalia/phpreport/web/services/getOpenTaskStoriesService.php on line 33, referer: http://localhost/igalia/phpreport/web/tasks.php?date=2016-07-11
`

tonythomas01:
Err. that seems to be unrelated. Looks like `newTask.id` is set to 1, when we click 'New' before load.

tonythomas01:
Alright. More findings:

- Task created soon do not have a `store:` value in it  (from console.log after creation), while others had the store value S 
- That would mean that `myStore.add(newTask)` is failing to add the same.

jaragunde:
Fixed by:
```
commit 5e4497de23eba879e905b0ff42458fc34c17b982
Author: Tony Thomas <01tonythomas@gmail.com>
Date:   Fri Jul 15 13:48:53 2016 +0530

    [#225] Wait till all tasks have loaded before adding in the new task
    
    This is a rude way of fixing the bug, but the right id seems not to
    load when the 'New Task' and a template is clicked before page load
```

Thank you!

jaragunde:
There were some JS errors due to this patch, solved with:
```
commit 495c0b6c427eb57bd25274ec33fc0be41056cfae
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Tue Nov 1 20:10:48 2016 +0100

    [#225] Prevent JS errors when creating tasks before everything is loaded
[...]

PR #258, opened 2016-07-13
[#225] Wait till all tasks have loaded before adding in the new task

tonythomas01:
This is a rude way of fixing the bug, but the right id seems not to
load when the 'New Task' is clicked before page load

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -85,6 +85,17 @@ var summaryRecord = new Ext.data.Record.create([
 /* Variable to store if there are unsaved changes */
 var unsavedChanges = false;
 
+/* Variable to store if all tasks has loaded completely for a day */
+var loaded = false;
+
+/**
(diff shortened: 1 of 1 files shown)


jaragunde:
First of all, thanks for the idea, I had never considered it :-)

The implementation is not taking into account the other ways to create tasks, like the [templates](https://github.com/Igalia/phpreport/blob/master/docs/user/tasks.rst#using-templates), which still have the same problem the "new task" button has.

I think you could either:
* Add a timeout for the template buttons too. Maybe you could refactor a bit the code for these two paths to share some more code.
* Disable the new task and all the template buttons while the load is not complete. Maybe you could reuse the code that disables them in the [block tasks](https://github.com/Igalia/phpreport/blob/master/docs/user/settings-management.rst#task-report-block-settings) feature.

tonythomas01:
I had to stick with the first plan. Do take a look

Issue #241, opened 2016-06-29
Project details top menu entry


jaragunde:
The discoverability of the project details report is not good. Let's add an entry for _Project details_ in the _Reports_ menu, which would open a page with a simple form with a project combo box. When a project is selected there, the project details page for it will open.

tonythomas01:
Addressed in #263

jaragunde:
After checking the implementation, I think it's necessary to add a client combo box too, which would filter the list of projects by client (like in the tasks screen).

jaragunde:
```
commit f72e752d595c73b3c4ee7d6b0ce304a17256fdba
Merge: e8fdc03 b16852a
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Thu Jul 21 17:06:00 2016 +0200

    [#241] Add 'Project Details' page to top menu entry
    
    Merge pull request #263 from tonythomas01/addTopMenu

commit b16852aef07df5ea96c68a4793d571d9de4dce76
Author: Tony Thomas <01tonythomas@gmail.com>
Date:   Thu Jul 21 14:02:55 2016 +0530

    [#241] Add filter by client to 'Project Details' page

commit 922359412620913f4dd42ce96fde36c1f7b0b987
Author: Tony Thomas <01tonythomas@gmail.com>
Date:   Tue Jul 19 23:42:52 2016 +0530

    [#241] Add 'Project Details' to top  menu entry
```

PR #263, opened 2016-07-19
[#241] Add 'Project Details' to top  menu entry

tonythomas01:
Added in the top menu as requested in the bug, and now takes the user to a new page with entire project details!

### DIFF
--- a/web/projectDetails.php
+++ b/web/projectDetails.php
@@ -1,0 +1,46 @@
+<?php
+/*
+ * Copyright (C) 2016 Igalia, S.L. <info@igalia.com>
+ *
+ * This file is part of PhpReport.
+ *
+ * PhpReport is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PhpReport is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


jaragunde:
The patch is looking good! Just some comments:
* When there is a lot of projects in the application, a client combo box becomes a must. Sorry for not realizing earlier.
* Please move the JavaScript code in projectDetails.php to a separate .js file, like in tasks.php and js/tasks.js. Many pages still mix php and js code but we want to get rid of that.
* I'm not sure if the details report should open in a new page, but let's keep it like that and let users decide.

jaragunde:
Ah, just another silly comment. Could you place the new menu entry above the "project evaluation" entry, and use lower case for the "details" word, like the other menu entries? :-)

web/projectDetails.php
```diff
@@ -0,0 +1,46 @@
...
+include_once("include/header.php");
+include_once("include/sidebar.php");
+include_once(PHPREPORT_ROOT . '/web/services/WebServicesFunctions.php');
+$user = $_SESSION['user'];
+?>
+
+<script type="text/javascript" src="js/include/ExportableGridPanel.js"></script>
+<script type="text/javascript">
+    var user = "<?php echo $user->getLogin(); ?>";
+</script>
```
jaragunde: I think this parameter doesn't make sense. We want to get the lists of customers and projects regardless of the user. Could you try to remove it?

tonythomas01:
I've rebased it over the other branch. Looks good ?

jaragunde:
Looks great now, thanks!

Issue #242, opened 2016-06-29
New report: tasks done by a person between two dates


jaragunde:
Description: The report will list all the tasks written in phpreport between two days. The filter would contain the following fields:

 * Worker
 * Start date time
 * End date time

The output will be all the tasks that the person has tracked between start date and end date. One line per task with the description field. It would be similar to _Reports -> User tasks_.

The new report to be added to the application top menu. The input fields for the report would be two dates and some user login.

jaragunde:
We could speed up the implementation of this page using the existing _User tasks report_ (`userTasksReport.php`) to implement it.
* We would add a new combo box to the page, where we would select which user we want to check. This combo should not be empty and should be pre-filled with the current user name.
* The backend service `getTasksFiltered.php` already allows to query the tasks of other users.
* We should refine the permissions on this page and service so:
  * Normal users: will find the users combo disabled. The backend service should return an error in case normal users query for another person's tasks.
  * Admin users: will find the users combo enabled.

To allow `userTasksReport.php` and `getTasksFiltered.php` to have different behaviors for normal and admin users, the currently unused "admin permission" infrastructure would be used. We can assign "admin permissions" on those pages to members of `admin` using the `$adminPermissions` array in `config/permissions.php`. The pages can check if users have that permission using `LoginManager::isAdmin()` and operate accordingly.

jaragunde:
> We could speed up the implementation of this page using the existing User tasks report (userTasksReport.php) to implement it.

To be more clear, let's not create a new report based on `userTasksReport.php`. Instead, let's add the users field to `userTasksReport.php` and add some restrictions on who can use it.

jaragunde:
```
commit 7cc144c47c6160e054a2c8195c2b491bca9db06f
Author: Tony Thomas <01tonythomas@gmail.com>
Date:   Fri Jul 22 21:11:47 2016 +0530

    [#242] New report: tasks done by a person between two dates
```

PR #264, opened 2016-07-22
[#242] New report: tasks done by a person between two dates

tonythomas01:
Right now the option gets activated only for admin users. Added checks in the backend service too, to avoid illegal use of the feature.


config/permissions.php
```diff
@@ -130,5 +130,5 @@
...
...iesService.php", "/services/deleteTaskStoriesService.php", "/services/updateTaskSectionsService.php", "/services/createTaskSectionsService.php", "/services/deleteTaskSectionsService.php", "/services/getSectionCustomTaskSectionsService.php", "/services/getAllUsersService.php", "/services/updateUsersService.php", "/services/createUsersService.php", "/services/deleteUsersService.php", "/services/getUserHourCostHistoriesService.php", "/services/getUserAreaHistoriesService.php", "/services/getUserCityHistoriesService.php", "/services/getUserJourneyHistoriesService.php", "/services/createJourneyHistoriesService.php", "/services/createHourCostHistoriesService.php", "/services/createAreaHistoriesService.php", "/services/createCityHistoriesService.php", "/services/updateJourneyHistoriesService.php", "/services/updateHourCostHistoriesService.php", "/services/updateAreaHistoriesService.php", "/services/updateCityHistoriesService.php", "/services/deleteJourneyHistoriesService.php", "/services/deleteHourCostHistoriesService.php", "/services/deleteAreaHistoriesService.php", "/services/deleteCityHistoriesService.php", "/viewProjects.php", "/userTasksReport.php", "/services/getTasksFiltered.php"),
```
jaragunde: Since we were not using this array at all, let's clear it completely and only add `"/userTasksReport.php", "/services/getTasksFiltered.php"`. Better if you add them with a comment saying "User tasks report" so we better know what it is for in the future.

web/js/userTasksReport.js
```diff
@@ -161,7 +197,33 @@ Ext.onReady(function () {
         width: 350,
         renderTo: 'content',
         defaults: {width: 230},
-        items: [{
+        items: [
+            new Ext.form.ComboBox({
```
jaragunde: An equivalent way to define items is creating an object which contains the correct `xtype`field, instead of creating instances with new. For example, you could have added an object with `xtype: 'combo'`. Take a look at the combo for customer, defined just some lines below.

web/js/userTasksReport.js
```diff
@@ -161,7 +197,33 @@ Ext.onReady(function () {
...
+                },
+                listeners: {
+                    'select': function () {
+                        this.baseParams.userId = this.getValue();
+                    },
+                    'blur': function () {
+                        if(this.baseParams.userId != this.getValue()){
+                            this.baseParams.userId = this.getValue();
+                        }
+                    }
```
jaragunde: You don't need these listeners here. Look at the implementation of the customers combo, for example; the value of the combo is retrieved later in `findTasks()`.

jaragunde:
Be also careful what happens when you remove the value of the user combo :) . Right now, it causes an error because it calls the service url like this:
`http://localhost/phpreport/web/services/getTasksFiltered.php?&userId=`.

Ideally, the combo box should be required to have some value, and the request should not be sent without that.

web/js/userTasksReport.js
```diff
@@ -285,12 +348,22 @@ Ext.onReady(function () {
...
     function findTasks () {
-                var baseParams = {
-                    'userId': userId,
-                };
+                var baseParams = {};
+
+                if (Ext.getCmp('userLogin').baseParams.userId == ""){
+                    // If no value was set in the combobox, default to the logged in user
+                    baseParams.userId = userId;
+                } else {
```
jaragunde: I agree this fixes the problem with sending an invalid request, but I still think it's better you set the proper flags to make the "user" combo compulsory, so users always have the feedback of whose tasks they are getting.

tonythomas01: True. I will just find the ext4 way of doing it :)

web/js/userTasksReport.js
```diff
@@ -285,12 +348,22 @@ Ext.onReady(function () {
...
+                    baseParams.userId = Ext.getCmp('userLogin').baseParams.userId;
```
jaragunde: I don't think you should be using the `baseParams` member of the combo box object to store the value. `baseParams` is intended to maintain a list of parameters that will be sent in every request to the server, and is a member of Store objects, not combo boxes.

It's better you use `Ext.getCmp('userLogin').getValue()` directly.

web/js/userTasksReport.js
jaragunde: Take a look at the time fields in the tasks screen, they are compulsory and tasks cannot be saved without them. It's a similar case.

Issue #243, opened 2016-07-04
Overtime control


jaragunde:
Include status information related to the target timetable for someone to fullfill for their acc hours for a defined period. Everyone could see how they would be doing in terms of acc hours reduction/increasing. The information would be present everytime you are tracking your time.

The task would imply to define a acc hours goal at a defined period, for each user. For example:

- 0acc hours at the end of the year, 
- 100acc hours at the end of the year.

The process would calculate the ideal timetable for each period of the month applying a lineal interpolation of hours. For instance, the ideal week timetable for someone could be 35.5 hours/week. So if in the current week, month or day someone has dedicated more or less than the estimated optimal scheduling, this person would see a -1h less or +25m more.

Different colors will be used: green or red depending on if someone is close to their goal.

jaragunde:
> The task would imply to define a acc hours goal at a defined period, for each user. For example:
> * 0acc hours at the end of the year,
> * 100acc hours at the end of the year.

I think it's simpler to implement, and even more useful, to define goals like this:
* in the period between 1st of January and 31st of December, I want to do 0 extra hours (I don't want to increase or reduce the total of accumulated hours).
* in the period between 1st of January and 31st of December, I want to do -50 extra hours (I want to reduce my total of accumulated hours by 50).

jaragunde:
This task is a bit complex, let's divide it in smaller parts. This would be the first step.

I would assume everyone's goal is 0 acc hours at between the first and the last day of the year. It means that the application should get these data:
* the number of _workable hours_ between the first and the last day of the year.
* the number of _worked hours_ since the first day of the year until the current date.

The data above can be extracted from the existing _Acc hours report_. Then, the application should:
* calculate the number of _hours left to be worked_, this is just: workable - worked.
* calculate the number of weeks since the current date until the end of the year.
* divide the hours left between the number of weeks. This is the number of hours the user should work per week to fulfill their goal at the end of the year.

The number of hours to be worked per week should be presented to the user in the _User work summary_ box.

tonythomas01:
I would start by creating a new menu entry under Reports -> My Goals, which can let the user set a from time, to time and extra hours he plan to work. This should go into a separate db table too where the extra time can take negative values too.

jaragunde:
I wouldn't add a new entry in the menu to define this, I think it would be enough to have a new tab in data management -> users -> user histories. You could reuse some classes like `BaseHistoryVO`.

jaragunde:
```
[...]

PR #274, opened 2016-07-30
[#243] Implement OverTime Control

tonythomas01:
More work required to actually implement the business logic, and show up in `tasks.php` sidebar

### DIFF
--- a/model/facade/action/GetPersonalSummaryByLoginDateAction.php
+++ b/model/facade/action/GetPersonalSummaryByLoginDateAction.php
@@ -90,6 +90,17 @@ class GetPersonalSummaryByLoginDateAction extends Action{
         $this->date = $date;
         $this->currentJourney = $dao->getByIntervals( $this->date, $this->date, $this->userVO->getId())[0];
         $this->currentUserGoal = $userGoaldao->getUserGoalsForCurrentDate( $this->userVO->getId(), $this->date );
+
+        if( !$this->currentUserGoal && $this->currentJourney ) {
+            $userGoalVO = new UserGoalVO();
+            $userGoalVO->setUserId( $userVO->getId() );
+            $userGoalVO->setExtraHours( 0 );
+            $userGoalVO->setInitDate( max( $this->currentJourney->getInitDate(),
+                DateTime::createFromFormat( 'Y-m-d', date('Y-m-d', strtotime('first day of January', $this->date->getTimestamp())))));
+            $userGoalVO->setEndDate( min( $this->currentJourney->getEndDate(),
+                DateTime::createFromFormat( 'Y-m-d', date('Y-m-d', strtotime('last day of December', $this->date->getTimestamp())))));
+            $this->currentUserGoal = $userGoalVO;
+        }
         $this->preActionParameter="GET_PERSONAL_SUMMARY_BY_USER_LOGIN_DATE_PREACTION";
         $this->postActionParameter="GET_PERSONAL_SUMMARY_BY_USER_LOGIN_DATE_POSTACTION";
 
@@ -103,10 +114,8 @@ class GetPersonalSummaryByLoginDateAction extends Action{
         if ( $this->currentUserGoal ) {
             $initDate = max($this->currentJourney->getInitDate(), $this->currentUserGoal->getInitDate());
             $endDate = min($this->currentJourney->getEndDate(), $this->currentUserGoal->getEndDate());
-        } else {
-            $initDate  = min($this->currentJourney->getInitDate(), DateTime::createFromFormat( 'Y-m-d', date('Y-m-d', strtotime('this week', $this->date->getTimestamp()))));
-            $endDate = $this->currentJourney->getEndDate();
         }
+
         //Now need to find out the workable hours in this year
         $extraHoursAction = new ExtraHoursReportAction($initDate, $endDate, $this->userVO);
         $results = $extraHoursAction->execute();
@@ -120,8 +129,6 @@ class GetPersonalSummaryByLoginDateAction extends Action{
     private function getWeeksTillEndOfJourneyPeriod() {
         if ( $this->currentUserGoal ) {
             $endDate = min($this->currentJourney->getEndDate(), $this->currentUserGoal->getEndDate());
-        } else {
-            $endDate = $this->currentJourney->getEndDate();
         }
 
         // Its always better to find the difference in weeks from the start of the week, rather than in between
--- a/model/facade/action/GetPersonalSummaryByLoginDateAction.php
+++ b/model/facade/action/GetPersonalSummaryByLoginDateAction.php
@@ -100,13 +100,13 @@ class GetPersonalSummaryByLoginDateAction extends Action{
      * @throws null
      */
     private function getWorkableHoursInThisJourneyPeriod() {
-        $initDate = $this->currentJourney->getInitDate();
         if ( $this->currentUserGoal ) {
+            $initDate = max($this->currentJourney->getInitDate(), $this->currentUserGoal->getInitDate());
             $endDate = min($this->currentJourney->getEndDate(), $this->currentUserGoal->getEndDate());
         } else {
+            $initDate  = min($this->currentJourney->getInitDate(), DateTime::createFromFormat( 'Y-m-d', date('Y-m-d', strtotime('this week', $this->date->getTimestamp()))));
             $endDate = $this->currentJourney->getEndDate();
         }
-
         //Now need to find out the workable hours in this year
         $extraHoursAction = new ExtraHoursReportAction($initDate, $endDate, $this->userVO);
         $results = $extraHoursAction->execute();
@@ -128,7 +128,6 @@ class GetPersonalSummaryByLoginDateAction extends Action{
         $thisWeekInitDay = DateTime::createFromFormat( 'Y-m-d', date('Y-m-d', strtotime('last sunday', $this->date->getTimestamp())));
         $lastWeekInitDay = DateTime::createFromFormat( 'Y-m-d', date('Y-m-d', strtotime('next monday', $endDate->getTimestamp())));
 
-
         $interval = $thisWeekInitDay->diff( $lastWeekInitDay );
         return floor($interval->days/7);
     }
@@ -140,7 +139,11 @@ class GetPersonalSummaryByLoginDateAction extends Action{
      */
     private function getWeeksInBetweenDates(DateTime $initDate, DateTime $endDate) {
         $interval = $initDate->diff( $endDate );
-        return floor($interval->days/7);
+        $weeksInBetween = floor($interval->days/7);
+        if ( $weeksInBetween == 0 ) {
+            return 1;
+        }
+        return $weeksInBetween;
     }
 
     /**
--- a/model/facade/action/GetPersonalSummaryByLoginDateAction.php
+++ b/model/facade/action/GetPersonalSummaryByLoginDateAction.php
@@ -30,6 +30,7 @@
  */
 
 include_once(PHPREPORT_ROOT . '/model/facade/action/Action.php');
+include_once(PHPREPORT_ROOT . '/model/facade/action/ExtraHoursReportAction.php');
 include_once(PHPREPORT_ROOT . '/model/dao/DAOFactory.php');
 include_once(PHPREPORT_ROOT . '/model/vo/UserVO.php');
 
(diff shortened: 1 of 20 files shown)


jaragunde:
In your first patch, watch out for:
* Commented debug code in the web services files.
* Copy-pasted comments in DAOs, Actions.
* Copyright years on new files.

In the second patch, there must be an error regarding how the weekly goals value is printed. My own 37 h figure (manually calculated) appears as 0:38. Also check some inline comments.

model/facade/action/GetPersonalSummaryByLoginDateAction.php
```diff
@@ -76,6 +77,42 @@ public function __construct(UserVO $userVO, DateTime $date) {
...
+     * @return mixed
+     * @throws null
+     */
+    private function getWorkedHoursInThisYear() {
+        $firstDayOfYear = date_create($this->date->format('Y-1-01'));
+
+        // We need to get the number of hours worked till date
+        $extraHoursAction = new ExtraHoursReportAction($firstDayOfYear, $this->date, $this->userVO);
+        $results = $extraHoursAction->execute();
+        return $results[1][$this->userVO->getLogin()]['total_hours'];
```
jaragunde: It's better if you modify ExtraHoursReportAction to take an optional $user parameter. That's a quite heavy report and you will be wasting resources if you calculate the report for every user and then discard them.

jaragunde: You are already doing that! My apologies :)

jaragunde:
Goals seem to work fine when they are specified in the 1st January - 31st Dec time period.

On the other hand, if you define goals in a different time period (I've tried to set 1st Aug - 4th Sep) it still works as if the end date was 31st Dec.

web/viewUsers.php
```diff
@@ -1047,6 +1050,146 @@ function loadHistories(login, userId){
...
+        id: 'goalGrid',
+        iconCls: 'silk-building',
+        width: 300,
+        store: goalStore,
+        frame: false,
+        header: false,
+        title: 'Goals',
+        loadMask: true,
+        colModel: goalColModel,
+        delMsg: 'Are you sure you want to delete the selected City History entries?',
```
jaragunde: Watch out for c&p issues like this :)

jaragunde:
I found a use case that doesn't look right. I've set a goal of -40h at the end of the year (Dec, 31st), and assigned a journey of 8h starting on Dec, 26th until Jan, 1st. That means 40h of work, to fulfil a goal of -40h it means the goal must be zero.

The results of the weekly goal vary depending on the date, like this:
* Monday 26th: 40h
* Tue-Fri: 0h
* Sat 31st: 40h

jaragunde:
Now I'm getting different results, but still strange ones. This is my setup:
* Journey: 8h between 1 Aug 2016 and 9 Aug 2017.
* City: some city with empty calendar between 1 Aug 2016 and 9 Aug 2017.
* Goals: 0h between 1 Aug 2016 and 31 Dec 2016.
* Empty user, no tasks saved.

With the setup above, my weekly goal is 39.42h for every week starting on week 31. This is wrong for several reasons:
* Goal on week 31 should be 40h.
* The weekly goal should grow every week we move into the future if the user hasn't added any tasks.

Additionally, there is a crash when jumping to any date earlier than the 1st of August:
```
[Wed Aug 10 18:57:01.467661 2016] [:error] [pid 20429] [client 127.0.0.1:55244] PHP Fatal error:  Call to a member function getInitDate() on null in /var/www/html/phpreport/model/facade/action/GetPersonalSummaryByLoginDateAction.php on line 95, referer: http://localhost/phpreport/web/tasks.php?date=2016-07-27

File: model/facade/action/GetWeeklyGoalSummaryAction.php [hunk 3]
```diff
@@ -0,0 +1,119 @@
...
+
+/*//Test code
+
+$user = new UserVO();
+$user->setLogin('jaragunde');
+$action = new GetPersonalSummaryByUserIdDateAction($user, date_create('2009-12-01'));
+//var_dump($action);
+$result = $action->execute();
+var_dump($result);
+*/
```
jaragunde: Remove this c&p test code, please :)

jaragunde:
One more thing, lately you've been amending your changes to the last commit of this set. That will make your patches more difficult to review. I recommend you keep working on the branch adding more patches to implement new aspects of the feature or correct issues in your implementation, except if they are trivial changes (like correcting a typo, renaming a variable, etc.).

tonythomas01:
The latest patch improves the accuracy a lot, and now in @jaragunde test case, I get 40.13 in the first week. That extra shows up during the `workable hours` calculation, not part of our code, though!

model/facade/action/GetPersonalSummaryByLoginDateAction.php
```diff
@@ -103,6 +103,17 @@ private function getWorkableHoursInThisJourneyPeriod() {
...
 
     /**
+     * @return float
+     */
+    private function getNumberOfWeeksInThisJourney() {
+        $initDate = $this->currentJourney->getInitDate();
+        $endDate = $this->currentJourney->getEndDate();
+
+        $interval = $initDate->diff( $endDate );
+        return round($interval->days/7, 1);
```
jaragunde: You are getting a float with one decimal digit, and I think you should get an integer. Shouldn't this be `ceil()`?

model/facade/action/GetPersonalSummaryByLoginDateAction.php
```diff
@@ -103,6 +103,17 @@ private function getWorkableHoursInThisJourneyPeriod() {
...
+    }
```
jaragunde: It looks like you aren't using this function!

model/facade/action/GetPersonalSummaryByLoginDateAction.php
```diff
@@ -122,13 +133,13 @@ private function getWorkedHoursInThisJourneyPeriod() {
     private function getWeeksTillEndOfJourneyPeriod() {
         $lastDayOfJourney = $this->currentJourney->getEndDate();
 
-        $interval = $this->currentJourney->getInitDate()->diff( $lastDayOfJourney );
-        return floor($interval->days/7);
+        $interval = $this->date->diff( $lastDayOfJourney );
+        return round($interval->days/7, 1);
```
jaragunde: Like I said before, I think the proper thing to do is using `ceil()`.

model/facade/action/GetWeeklyGoalSummaryAction.php
```diff
@@ -106,14 +106,4 @@ protected function doExecute() {
...
-/*//Test code
-
-$user = new UserVO();
-$user->setLogin('jaragunde');
-$action = new GetPersonalSummaryByUserIdDateAction($user, date_create('2009-12-01'));
-//var_dump($action);
-$result = $action->execute();
-var_dump($result);
-*/
+}
```
jaragunde: This is the kind of trivial change I would have amended into the previous commit O:-)

No need to do it now, just take it into account in the future.

jaragunde:
I've retaken one of the use cases from above: I've assigned a journey of 8h starting on Dec, 26th until Jan, 1st. That means 40h of work on week 52 of year 2016.

Now these are the results of the weekly goal for that week:
* Monday: 44:26
* Tuesday: 57:08
* (it keeps increasing)
* Friday: 133:19
* Saturday: 400!

The problem with different values for different days in the same week is probably related with the way you round week calculations, you should be always using `ceil()`.

model/facade/action/GetPersonalSummaryByLoginDateAction.php
```diff
@@ -69,13 +78,69 @@ class GetPersonalSummaryByLoginDateAction extends Action{
...
+    }
+
+    /**
+     * @return float
+     */
+    private function getWeeksTillEndOfJourneyPeriod() {
+        $lastDayOfJourney = $this->currentJourney->getEndDate();
+
+        $interval = $this->date->diff( $lastDayOfJourney );
+        return floor($interval->days/7);
```
jaragunde: I'm getting a 0 weekly goal in the w52 example. If I change this `floor` with `ceil`, it works properly.

model/facade/action/GetPersonalSummaryByLoginDateAction.php
```diff
@@ -69,13 +78,69 @@ class GetPersonalSummaryByLoginDateAction extends Action{
...
+    /**
+     * @return float
+     */
+    private function getNumberOfWeeksInThisJourney() {
+        $initDate = $this->currentJourney->getInitDate();
+        $endDate = $this->currentJourney->getEndDate();
+
+        $interval = $initDate->diff( $endDate );
+        return ceil($interval->days/7);
+    }
```
jaragunde: I can't see this function used anywhere! If you don't need it, please remove it.

jaragunde:
We are getting closer! But there is still a dependency on the journey end date that should not be there. Take [this case](https://github.com/Igalia/phpreport/pull/274#issuecomment-238931747) again as an example. It's still exposing different results depending on the day of the week:

* Tuesday Aug, 2nd: 39:42 h
* Wednesday Aug, 3rd: 40:27 h

Ideally, the result for any day in that week should be 40h.

jaragunde:
Getting closer and closer, but I just noticed something else O:-)

Using [this case](https://github.com/Igalia/phpreport/pull/274#issuecomment-238931747) again. Now the results on week 31 are 40:27 h for every day in the week. As I said before, the result for any day in that week should be 40h. What's more, the value still changes if I move the end date of the journey. If I set the journey end date to 31st Dec 2016, the weekly goal becomes 41:54 h for week 31. The goal was set for December, 31st, the change of the journey end date beyond 2016 should not affect the result.

jaragunde:
We are almost there, now we have an issue when you start entering your hours :)

Using [this case](https://github.com/Igalia/phpreport/pull/274#issuecomment-238931747) again. Enter Aug, 1 and you will see the weekly goal is 40h, that's correct. Now add a task with 8 work hours; when you save it the weekly goal will go down to 39:38 h.

The tasks I enter on the current week should not affect my goal for it. The goal should still say 40h so I know I need to have 40h at the end of my work week on Friday.

jaragunde:
I still can reproduce the problem I just mentioned. BTW, I don't get how the worked hours affect your calculations, I must be missing something in your code :)

jaragunde:
> I still can reproduce the problem I just mentioned.

It was a problem on my side! Now I'm sure I'm running the latest version, and it works.

> BTW, I don't understand how the worked hours affect your calculations, I must be missing something in your code :)

Now I see worked hours don't affect your calculations, but that's a problem because the goal doesn't get updated as you work.

Imagine a very simple case, with a user who has a contract between Aug 15 and Aug 28, and their journey is 8h. That means they have two weeks of work, 40h each. The goal for the first week is 40h and the goal for the second week is 80h.

Now imagine that user works normally on the first week, and fulfils the goal of 40h. The goal for the second week should be 40h, but the application still says it's 80.

tonythomas01:
The latest patch solves the exact problem!

jaragunde:
I found a new corner case, although I admit it's a strange one. This is the situation:
* Journey is 8h between Aug 15 and 28.
* Assigned to some city between those dates too.
* Goal is set to -40h between Aug 15 and 21 (not 28!).

This is how I think it should work:
* On week 33, there are 40 workable hours and goal is -40 at the end of that week. That means the weekly goal should be 0, but it is 40 instead.
* On week 34 there are 40 remaining workable hours and default goals apply, which means it should be 0 at the end of the contract.  That means the weekly goal should be 40, but it is 80 instead.

Keep up the good work! This is almost done.

jaragunde:
When trying to fix the situation above, take into account this scenario: https://github.com/Igalia/phpreport/pull/274#issuecomment-238931747

If you do changes to how the init date is calculated, take into account that scenario should work exactly the same when you remove the goal.

jaragunde:
Regarding this scenario: https://github.com/Igalia/phpreport/pull/274#issuecomment-239465993

I think we could reconsider this situation:

> On week 34 there are 40 remaining workable hours and default goals apply, which means it should be 0 at the end of the contract. That means the weekly goal should be 40, but it is 80 instead.

We could leave the result as 80, and assume that when a goal is over, it resets to the default periods. Currently those are:
* init date: min(journey start, current week)
* end date: journey end

I don't really like to use journey start and end dates if they span longer than the current year, so I would use Jan 1, Dec 31 as limits.

Issue #246, opened 2016-07-04
Autosave tasks


jaragunde:
Description: Nowadays you need to press the Save button when you create a task. The idea will be to auto-save the data when there are changes in the UI fields.

The auto-save will be run every certain amount of time (we will need to adjust which one works better) and will have to check if there are errors in any fields before doing it. The save operation will not be performed until all tasks in the page have been correctly validated.

There should be a visual indication for users to know if there are unsaved changes or not, so they can be sure before closing the tab.

Rationale: The need to press the save button is an overhead and it causes also errors when someone forgets to press it.

jaragunde:
```
commit ab59758a4b3965901306299a04fa5eae39f4c0cd
Author: Tony Thomas <01tonythomas@gmail.com>
Date:   Mon Aug 1 21:00:42 2016 +0530

    [#246] Autosave tasks, once valid
    
    TODO:
    * Would look more good if the text is moved to the right end.
```

jaragunde:
The "ok" banner should not appear when tasks are autosaved, it's a bit annoying to have it appearing every 10 seconds while you type your tasks in.

jaragunde:
```
commit 7658175893c26fcb9f845159d4cf22c835a3080c
Author: Tony Thomas <01tonythomas@gmail.com>
Date:   Mon Aug 22 11:02:17 2016 +0530

    [#246] Show up 'saved-task' App banner only on manually saved tasks
```

jaragunde:
Detected regression.

jaragunde:
```
commit 539ee9bff1151c79f9d466a8734fd9ee4b2288a7
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Mon Oct 10 14:22:29 2016 +0200

    [#246] Show up 'saved-task' App banner only on manually saved tasks
    
    Re-applied commit 7658175, due to regression caused by 3855718.

PR #276, opened 2016-08-01
[#246] Autosave tasks, once valid

tonythomas01:
TODO:
* Would look more good if the text is moved to the right end.

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -804,6 +804,7 @@ Ext.onReady(function(){
             },
             'save': function () {
                 if (!myStore.error) {
+                    Ext.getCmp('status_display').setText("Status: saved at "+ new Date());
                     App.setAlert(true, "Task Records Changes Saved");
                     summaryStore.load();
                     unsavedChanges = false;
@@ -816,9 +817,11 @@ Ext.onReady(function(){
             },
             'update': function () {
                 unsavedChanges = true;
+                Ext.getCmp('status_display').setText("Status: Pending changes");
             },
             'remove': function () {
                 unsavedChanges = true;
+                Ext.getCmp('status_display').setText("Status: Pending changes requiring manual save");
             }
         }
     });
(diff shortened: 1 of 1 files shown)


web/js/tasks.js
```diff
@@ -884,6 +884,37 @@ Ext.onReady(function(){
...
+            var valids = true;
+            for(var panel=0; panel<panels.getCount(); panel++) {
+                console.log();
+                if (!panels.get(panel).initTimeField.isValid() ||
+                    !panels.get(panel).endTimeField.isValid() ||
+                    !panels.get(panel).customerComboBox.getRawValue() ||
+                    !panels.get(panel).projectComboBox.getRawValue() ||
+                    !panels.get(panel).storyField.getRawValue() ||
+                    !panels.get(panel).descriptionTextArea.getRawValue()
+                ) {
```
jaragunde: The only compulsory fields are init and end time. Tasks may be saved with empty customer, project, etc.

jaragunde:
Have you tried to set `autosave: true` in the tasks Store? Maybe it works...

tonythomas01:
`autosave:true` makes it save on every change (even submitting when the data is not entered, throwing up an app error ).

web/js/tasks.js
jaragunde: Hmmm, it's better not to duplicate the validation code. It already lives in the `saveTasks()` function. Maybe you could split that method in two parts (validateTasks and saveTasks) to make it easier to reuse.

web/js/tasks.js
```diff
@@ -884,6 +884,37 @@ Ext.onReady(function(){
...
+            }
+            if (valids) {
+                myStore.each(function(r) {
+                    if (r.data['story'] != undefined)
+                        r.data['story'] = xmlencode(r.data['story']);
+                    if (r.data['text'] != undefined)
+                        r.data['text'] = xmlencode(r.data['text']);
+                });
+                myStore.save();
+                unsavedChanges = false;
```
jaragunde: Don't duplicate this code either, better rework `saveTasks()` a little bit to make it work as you want so you may call it from here.

Also, `unsavedChanges = false` should not happen here, only after the store is successfully saved.

web/js/tasks.js
```diff
@@ -884,6 +884,37 @@ Ext.onReady(function(){
...
+                Ext.getCmp('status_display').setText("Status: Auto saved at "+ new Date());
```
jaragunde: It's not really important if the tasks were manually or automatically saved. You may move this message to the save event in the Store object.

web/js/tasks.js
```diff
@@ -882,34 +883,53 @@ Ext.onReady(function(){
...
+        // First we check if the time fields of all records are valid, and then save
+        if (validateTasks()) {
+            addToMyStore();
         } else  // Otherwise, we print the error message
           App.setAlert(false, "Check For Invalid Field Values");
     }
 
+    // Implement autosave of data, when valid contents are typed in to the task fields
+    window.setInterval(function () {
+        if( unsavedChanges == true ) {
```
jaragunde: Don't compare a boolean like this, you could have written `if(unsavedChanges)....`. In this case, you may even use `if(isUnsaved())`

jaragunde:
Hi Tony, this is looking great now. I've just noticed that the *status* indicator does not notice when you have made changes to a task. For example, create a task, save it and then modify its description. 

Another situation is when you open a date that has some saved tasks, the indicator says "status: draft" even when you haven't done any changes yet. Maybe it should say "unchanged".

I think you could implement the status indicator with a callback that triggers periodically and checks the status of `isUnsaved()`. This function already takes into account changes inside the tasks. Maybe you could reuse the same function that triggers the save operation.

jaragunde:
I noticed an error in `isUnsaved()` implementation when testing your patch. Please rebase to the latest master :)

jaragunde:
> I think you could implement the status indicator with a callback that triggers periodically and checks the status of isUnsaved().

Or you could modify the value of _status_ in the `update`, `remove` events of the store.

Issue #65, opened 2016-03-09
Autocomplete doesn't work when PhpReport is left opened for a long time


jaragunde:
[Trac import]
Reported by: jaragunde
Original date: Wednesday, 07 April 2010 17:25

The cause of the problem could be that sessions expire after a while, but it has to be double-checked if that's the problem.

It's a bit annoying to reproduce because the application has to be left opened for a long time. Is there any way to shorten that time with testing purposes?

jaragunde:
[Trac import]
Comment by: jaragunde
Original date: Thursday, 15 April 2010 10:18

The bug has an additional problem, after the application being opened for a long time, you cannot save the tasks and you have to reload, losing the changes.

I'm rising the priority because the usual behaviour of users is leaving the application opened aside while they're working, causing this problem.

jaragunde:
[Trac import]
Comment by: jaragunde
Original date: Thursday, 15 April 2010 10:19

Óscar has made me know that he could reproduce the problem in Firefox but not in Chrome.

jaragunde:
[Trac import]
Comment by: aperez
Original date: Wednesday, 09 June 2010 12:52

I can confirm that this happens in Chromium. I left !PhpReport open, suspended the computer to RAM, went to sleep, and when I tried to use it (approx. after ~8 hours) I was not able of use the autocompletion.

jaragunde:
[Trac import]
Comment by: jaragunde
Original date: Tuesday, 16 November 2010 14:17

I've cheched that the session is closed after two hours, maybe even in less time.

If you try to save after having left the browser open for a long time, the web service returns the following answer:
```
<?xml version="1.0"?>
<return service="createTasks"><success>false</success><error id="2">You must be logged in</error></return>
```

So I deduce two posible causes: the session is closed by the PHP-side of the application, or the cookie storing the session id on the browser expires.

jaragunde:
[Trac import]
Comment by: jaragunde
Original date: Wednesday, 24 November 2010 19:06

After some testing, I've gathered some facts:
   * I was able to reproduce the bug waiting one hour on a server with HTTP authentication enabled, but I wasn't able to do that in the same time in my local machine, which doesn't use HTTP auth (not sure it affects).
   * The cause of the bug is that the ```$_SESSION``` variable is deleted on server side, so the method ```LoginManager::isLogged()``` doesn't find ```$_SESSION["user"]``` and it returns false.
[...]

PR #278, opened 2016-08-09
[#65] Prompt user to login again when session has expired in background

tonythomas01:
Can be tested by logging out from another tab by keeping the session active in the main one.

### DIFF
--- a/web/services/checkSessionActive.php
+++ b/web/services/checkSessionActive.php
@@ -1,0 +1,35 @@
+<?php
+/*
+ * Copyright (C) 2016 Igalia, S.L. <info@igalia.com>
+ *
+ * This file is part of PhpReport.
+ *
+ * PhpReport is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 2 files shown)


web/services/checkSessionActive.php
```diff
@@ -0,0 +1,37 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PhpReport.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** login web service.
+ *
+ * @filesource
+ * @package PhpReport
```
jaragunde: There is some trailing whitespace at the end of this line O:-)

web/services/checkSessionActive.php
```diff
@@ -0,0 +1,37 @@
...
+/* We check authentication and authorization */
+require_once(PHPREPORT_ROOT . '/util/LoginManager.php');
+
+if (LoginManager::isLogged()) {
+	http_response_code(200);
+} else {
+	http_response_code(401);
+}
+
+
```
jaragunde: Now you are at it, remove the second while line in the end of the file.

PR #280, opened 2016-08-11
[#244] Introduce new 'manager' profile, restrict 'staff' access

### LINKED ISSUE #244: Create new "manager" profile
Now the standard users have access to all the information stored in PhpReport through the different reports, but we would like to have more granularity.

Let's create a new profile (user group) called _manager_ which would have access to all kinds of reports, and limit the capabilities of the _user_ profile.

The _user_ profile will only be able to:
* create and edit their own tasks.
* check their own hours (user tasks and user details reports).
* check the accumulated hours report.
* check the project details of the projects they are assigned to.

tonythomas01:
Fix #244

### DIFF
--- a/config/permissions.php
+++ b/config/permissions.php
@@ -125,7 +125,7 @@ $permissions = array(
         '/settings.php')
 );
 
-/** Admin permissions array
+/** Extra permissions array
  *
  * It contains a multiple-level array with the groups and the
  * pages they are allowed to act on as admin. In every element of
@@ -133,10 +133,16 @@ $permissions = array(
  * another array with the urls the group is allowed to open.
  * For example: "clients" => array("/xptracker-summary.php")
  */
-$adminPermissions = array(
+$extraPermissions = array(
     "admin" => array(
         //user tasks report
         "/userTasksReport.php", "/services/getTasksFiltered.php",
         //users management
-        "/viewUsers.php"),
+        "/viewUsers.php",
+    ),
+    'manager' => array(
+        '/viewProjectDetails.php', '/projectDetails.php', '/viewUserDetails.php',
+        '/services/getProjectUserCustomerReportJsonService.php',
+        '/services/getUserProjectCustomerReportJsonService.php'
+    )
 );
--- a/config/permissions.php
+++ b/config/permissions.php
@@ -56,19 +56,15 @@ $permissions = array(
         '/services/getProjectAnalysisTrackerTree.php', '/services/getProjectTrackerTree.php',
         '/services/getStoryCustomTaskStoriesService.php', '/services/getSectionCustomTaskSectionsService.php',
         '/services/updateTaskStoriesService.php', '/services/createTaskStoriesService.php',
-        '/services/deleteTaskStoriesService.php',
-        '/services/updateTaskSectionsService.php', '/services/createTaskSectionsService.php',
-        '/services/deleteTaskSectionsService.php',
+        '/services/deleteTaskStoriesService.php', '/services/updateTaskSectionsService.php',
+        '/services/createTaskSectionsService.php', '/services/deleteTaskSectionsService.php',
         //reports
-        '/viewProjectDetails.php', '/viewUserDetails.php', '/projectsSummary.php', '/projectDetails.php',
-        '/viewWorkingHoursResultsReport.php', '/projectsEvaluation.php', '/usersEvaluation.php',
-        '/services/getExtraHoursReportService.php', '/services/getPendingHolidayHoursService.php',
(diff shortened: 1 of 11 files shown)


config/permissions.php
```diff
@@ -82,6 +82,9 @@
         '/services/getUserCityHistoriesService.php', '/services/getUserJourneyHistoriesService.php',
         '/services/getTodayAreaUsersService.php', '/services/getProjectUsersService.php',
         '/services/getProjectCustomersService.php'),
+    'manager' => array(
+        '/projectsEvaluation.php', '/usersEvaluation.php', '/projectsSummary.php',
+    ),
```
jaragunde: You need to check which are the services related to those report screens and prevent access to them too.

sql/update/add-manager-user-group.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO user_group VALUES (3,'manager');
```
jaragunde: Please, add a small explanation comment on this file, like in other update files.

jaragunde:
The patch is looking good! I left you some comments in the code.

Additionally, there is a missing part of the task, do it as a new patch on top of this one:
> The user profile will only be able to:
> ...
> * check the project details of the projects they are assigned to.

You need to add a check for `isAdmin()` in that report; only `manager` and `admin` should have admin in that screen. If user has not admin, it should only be able to see the details page for the projects it is assigned to. There is some facade operation to retrieve the list of users assigned to a project.

tonythomas01:
I will introduce the second feature as another commit in the PR!

jaragunde:
The patch looks good, but you forgot to let the manager profile check the details of any project, as specified before:
> only manager and admin should have admin in that screen. 

Let's also make the next change as part of this task (in a separate commit):

> isAdmin() doesn't sound right because there already is a user group called admin. Maybe hasExtraPermissions() is a better name, feel free to suggest one.

config/permissions.php
```diff
@@ -81,7 +81,17 @@
...
+        //reports
+        '/viewProjectDetails.php', '/viewUserDetails.php', '/projectDetails.php', '/viewWorkingHoursResultsReport.php',
+        '/services/getExtraHoursReportService.php','/services/getPendingHolidayHoursService.php',
+        '/services/getProjectTtypeReportService.php', '/services/getProjectUserCustomerReportJsonService.php',
+        '/services/getProjectUserCustomerReportService.php', '/services/getProjectUserStoryReportService.php',
+        '/services/getUserProjectCustomerReportJsonService.php', '/services/getUsersProjectsReportService.php',
+        '/services/getProjectUserStoryReportJsonService.php',
+        '/services/getProjectUserWeeklyHoursReportJsonService.php', '/services/getUserCustomersService.php',
+        '/services/getUserProjectCustomerReportJsonService.php'
+
```
jaragunde: As things are now, an admin must be also member of staff for things to work correctly. It's reasonable to consider that a manager should also be staff, so you don't have to repeat the full list of report pages here.

So it was correct before the change :)

config/permissions.php
```diff
@@ -125,18 +135,24 @@
...
-$adminPermissions = array(
+$extraPermissions = array(
     "admin" => array(
         //user tasks report
-        "/userTasksReport.php", "/services/getTasksFiltered.php",
+        "/userTasksReport.php", "/services/getTasksFiltered.php", "/viewUserDetails.php",
         //users management
-        "/viewUsers.php"),
-);
+        "/viewUsers.php", "/services/getUserProjectCustomerReportJsonService.php"
```
jaragunde: Same here! An admin needs to be a manager too, so you don't need to repeat `viewUserDetails` and `getUserProjectCustomerReportJsonService`.

config/permissions.php
```diff
@@ -81,7 +81,7 @@
     'manager' => array(
         '/projectsEvaluation.php', '/usersEvaluation.php', '/projectsSummary.php', '/services/getFilteredCustomProjectsService.php',
         '/services/getUserStoryReportJsonService.php', '/services/getProjectCustomerReportJsonService.php',
-        '/services/getProjectUserReportJsonService.php',
+        '/services/getProjectUserReportJsonService.php', '/services/getCustomerProjectsService.php',
```
jaragunde: `getCustomerProjectsService` is already listed in `staff`, you don't need it here.

jaragunde:
I would have preferred to have the isAdmin -> hasExtraPermissions on a separate commit, but that's fine now :)

As a rule of thumb, if you need a separate line in the commit message to explain some part of your changes, they probably belong to a different commit.

PR #282, opened 2016-08-13
[#65] Improved session checks on all pages, reduce ping to once every…

### LINKED ISSUE #65: Autocomplete doesn't work when PhpReport is left opened for a long time

tonythomas01:
Fix #65

### DIFF
--- a/web/analysistracker-summary.php
+++ b/web/analysistracker-summary.php
@@ -39,7 +39,7 @@ include("include/sidebar.php");
 ?>
 
 <link rel="stylesheet" type="text/css" href="include/ColumnNodeUI.css" />
-
+<script src="js/include/sessionTracker.js"></script>
 <script type="text/javascript" src="include/ColumnNodeUI.js"></script>
 <script type="text/javascript" src="include/AnalysisTrackerSummaryTree.js"></script>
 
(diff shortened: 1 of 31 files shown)


jaragunde:
Looks good! Just some comments:
* Don't forget the copyright & license header in the new file.
* Could you please move `web/include/sessionTracker.js` to `web/js/include`? Sorry it's not mentioned anywhere, we are trying to put any new JavaScript code in that location.

And now you have to push a new version of the patch, could you please rebase it on top of master?

jaragunde:
Do you know the `sed` command? This will help you change the file path everywhere:
```
sed -i 's_include/sessionTracker.js_js/include/sessionTracker.js_' *.php

tonythomas01:
Wow. The `sed` worked like a charm. I always wanted to try that out once, and this was the time :) Thank you! :+1:

Issue #271, opened 2016-07-29
Combined search in combo box for customers and projects


jaragunde:
When adding a task, the customer field is usually just used as a filter for the project field. It would be ideal we had only one combo box where users could type project or customer names to get a filtered list of projects.

It could look like this:

```

--------------------------------
| Goog                      |\/|
--------------------------------
 |  Chromium project - Google  |
 |  Hangouts project - Google  |
 -------------------------------
 
--------------------------------
| Ch                        |\/|
--------------------------------
 |  Chromium project - Google  |
 |  Chakra project - Microsoft |
 -------------------------------
```

jaragunde:
I found this example in Sencha forums, it should work for our purpose: http://extjs.com/forum/showthread.php?p=209279#post209279

jaragunde:
Fixed in the same patch set as #248:
```
```
65c0490 (HEAD -> master, origin/master, origin/HEAD) [#248] [#271] Remove "client" as a compulsory field for tasks [GitHub]
d16c4fa (tony/remove-client) [#248] Remove customer relations completely from tasks and userTasksReports [Tony Thomas]
03d9055 [#248] Replace Project-Customer report in viewUserDetails with Project-Worked hours [Tony Thomas]
849028d [#248] Do not populate task/template table with customerid anymore [Tony Thomas]
a28cf41 [#248][Cleanup] Remove use of RequestDAO from codebase [Tony Thomas]
bdb2a60 [#248] Remove customer-project relations from viewProjects.php [Tony Thomas]
ed2e8d9 [#248] Allow filter via projectname-customer name in userTasksReport.php [Tony Thomas]
e3ecb60 [#248] Show only active projects on taks page and the full list on viewPorjects [Tony Thomas]
36b04ef [#248] Remove the project/customer report in the projects summary page [Tony Thomas]
e5f8866 [#248] Replace user-customer grid with total hours for all users for a project [Tony Thomas]
34beaa5 [#248] Allow filter via projectname-customer name in projectDetails.php [Tony Thomas]
6bb6325 [#248] Disable customercombobox, make it show up below project combo [Tony Thomas]
771ae87 [#271] Populate the task screen and database with the correct customer details [Tony Thomas]
cf5a5c2 [#271] Show projects without a client assigned too in the tasks screen [Tony Thomas]
3855718 [#271] Show up customers along with projects in tasks screen [Tony Thomas]
3147d3d [#248] Migration script to populate customerId in projects table [Tony Thomas]
[...]

PR #288, opened 2016-08-23
[#248] Remove "client" as a compulsory field for tasks

### LINKED ISSUE #248: Remove "client" as a compulsory field for tasks
Description: At present it is needed to fill always the client when you track a task in phpreport. This is so because a project can belong to several customers in the current data model. However, that is not a case to take into account. It does only happen in a few cases and can be changed how this was filled. Therefore, one project should belong only to a customer. If this is this way, it is enough to choose the project in the task to deduce the customer. This feature consists of changing the data model so that a project can have only a customer and, then, to change the UI so that if you fill the project the save operation is successful.

Rationale: It provides the following advantages:
* It simplifies the data model of the database.
* It simplies and makes faster to fill the phpreport.
* It saves space in the screen.

Assessment: It is needed to solve manually any existing projects which are assigned to several customers. A script should be developed to change this: it should assign the project to the most frequent customer, leave old customer as an unused column in tasks table.

tonythomas01:
Fix #248


sql/update/populate_customer_in_projects.php
```diff
@@ -0,0 +1,78 @@
...
+ * This script selects the customer id related to each project and populate the new customerid
+ * column in project table.
+ */
+
+define('DB_NAME', 'phpreport');
+define('DB_USER', 'phpreport');
+define('DB_PASSWORD', 'phpreport');
+define('DB_HOST', 'localhost');
+define('DB_PORT', 5432);
+
```
jaragunde: I see you used `sql/anonymize.php` as an example, but I expected you to use `update/update-from-2.1-to-2.16.php` and the like. Take a look at those; they include these values from config.php.

Also, please move this script to that directory. I'll probably rename it as `update/update-from-2.16-to-2.17.php` at some point.

sql/update/populate_customer_in_projects.php
```diff
@@ -0,0 +1,78 @@
...
+	" dbname=".DB_NAME.
+	" password=".DB_PASSWORD);
+if (!$link) error_log("ERROR: Could not connect to DB");
+
+$result = pg_query($link, $query = "SELECT customerid, projectid FROM requests");
+if ($result == NULL) error_log("ERROR: Could not run query: $query");
+$frequency = array();
+
+while ($row=pg_fetch_array($result, NULL, PGSQL_ASSOC)) {
+	$frequency[$row['projectid']][$row['customerid']] = true;
```
jaragunde: `$frequency` is not a good name for this variable, it has no relation with its contents. Also, you are using the keys to contain the customer ids, and discarding the content of the variable, but I think it's cleaner to use arrays like this:
```
$frequency[$row['projectid']][] = $row['customerid'];
```
It will make your array look like:
```
  [256]=>
  array(1) {
    [0]=>
    string(2) "33"
  }
  [257]=>
  array(17) {
    [0]=>
    string(2) "33"
    [1]=>
    string(2) "39"
    [2]=>
    string(2) "34"
    [3]=>
    string(2) "41"
    [4]=>
    string(1) "4"
    [5]=>
    string(2) "16"
    [6]=>
    string(2) "19"
    [7]=>
    string(2) "35"
    [8]=>
    string(2) "31"
    [9]=>
    string(2) "11"
    [10]=>
    string(2) "38"
    [11]=>
    string(1) "5"
    [12]=>
    string(2) "42"
    [13]=>
    string(2) "10"
    [14]=>
    string(1) "7"
    [15]=>
    string(1) "9"
    [16]=>
    string(2) "14"
  }

File: sql/update/populate_customer_in_projects.php [hunk 3]
```diff
@@ -0,0 +1,78 @@
...
+}
+
+$projectCustomerFreq = array();
```
jaragunde: You don't really need this array, as you don't need to store the values outside the foreach loop.

sql/update/populate_customer_in_projects.php
```diff
@@ -0,0 +1,78 @@
...
+foreach ( $frequency as $projectId => $value ) {
+	$projectCustomerFreq[$projectId] = key($value);
```
jaragunde: I'm not a big fan of this name either, it doesn't have to do with frequency at this point. Also, it doesn't need to be an array, as I said before; it's enough with a variable that is overwritten for every loop.

sql/update/populate_customer_in_projects.php
```diff
@@ -0,0 +1,78 @@
...
+			$result = pg_query( $link, $query="select count(*) as freq from task where projectid = $projectId and customerid = $customerId" );
+			$row = pg_fetch_assoc( $result );
+			$taskCustomerFreq[$projectId][$customerId] = $row['freq'];
+		}
+
+		//We sort by value to get the highest number of customer association
+		foreach ($taskCustomerFreq as $projectFreq ) {
+			arsort($projectFreq);
+			$taskCustomerFreq[$projectId] =  $projectFreq;
+		}
```
jaragunde: Please take into account a situation where the project has no assigned tasks, or the assigned tasks contain empty values for `customerId`, it could happen. Please test those situations creating some new projects with these conditions.

sql/update/populate_customer_in_projects.php
```diff
@@ -0,0 +1,78 @@
...
+
+$projectCustomerFreq = array();
+foreach ( $frequency as $projectId => $value ) {
+	$projectCustomerFreq[$projectId] = key($value);
+	if ( count( $frequency[$projectId]) > 1 ) {
+		$taskCustomerFreq = array();
+		foreach ($value as $customerId => $dummy ) {
+			// We have more than one customers for these projects, so lets check
+			// in tasks table
+			$result = pg_query( $link, $query="select count(*) as freq from task where projectid = $projectId and customerid = $customerId" );
```
jaragunde: You can use a simpler query! Try this:
```
select customerId, count(*) from task where projectId=$projectId group by customerId;

File: update/update-from-2.16-to-2.17.php [hunk 7]
```diff
@@ -0,0 +1,170 @@
...
+
+
+require_once(PHPREPORT_ROOT . 'config/config.php');
+include_once(PHPREPORT_ROOT . '/util/DBPostgres.php');
+
+if (!check_db_version(DB_HOST,DB_PORT,DB_NAME,DB_USER,DB_PASSWORD)) {
+	print ("Wrong database version. " .
+		"Make sure DB is on 2.1 version before running this upgrade.\n");
+	exit();
+}
```
jaragunde: The code right above here must be a copy&paste error, right?

update/update-from-2.16-to-2.17.php
```diff
@@ -0,0 +1,170 @@
...
+
+	$customerAssignedToProject = array();
+
+	while ( $row = pg_fetch_array( $result, null, PGSQL_ASSOC ) ) {
+		$customerAssignedToProject[$row['projectid']][] = $row['customerid'];
+	}
+
+	foreach ( $customerAssignedToProject as $projectId => $value ) {
+		$customerAssociated = $value[0];
+		if ( count( $customerAssignedToProject[$projectId] ) > 1 ) {
```
jaragunde: You can probably use `count($value)` here. Even, instead of `$value`, you could call it `$customerList` for the code to be more readable.

update/update-from-2.16-to-2.17.php
```diff
@@ -0,0 +1,170 @@
...
+			$taskCustomerFreq = array();
+			foreach ( $value as $customerId => $dummy ) {
```
jaragunde: Now you are not saving the customer ids as keys in the array! This should be `foreach ($value as $customerId)`.

update/update-from-2.16-to-2.17.php
```diff
@@ -0,0 +1,170 @@
...
+			if ( isset( $taskCustomerFreq[$projectId] ) ) {
+				$customerAssociated = key( $taskCustomerFreq[$projectId] );
+			}
+		}
+		if ($customerAssociated) {
+			$update_query = "UPDATE project SET customerid= $customerAssociated where id= $projectId";
+			print( $update_query . "\n" );
+			if ( !pg_query( $link, $update_query ) ) {
+				error_log( "ERROR: query failed: $update_query" );
+			}
```
jaragunde: You are mixing code styles in this file, be careful with `if ( $customerAssociated )` and `if ($customerAssociated)` and the like. I prefer the latter, which is more widely used in the code base.

update/update-from-2.16-to-2.17.php
```diff
@@ -0,0 +1,170 @@
...
+		}
+	}
+
+	pg_freeresult( $result );
+}
+
```
jaragunde: I've just noticed this file is indented with tabs instead of spaces.

update/update-from-2.16-to-2.17.php
```diff
@@ -0,0 +1,167 @@
...
+            foreach ($customerList as $customerId) {
+                // We have more than one customers for these projects, so lets check
+                // in tasks table
+                $result = pg_query($link, $query = "select customerid, count(*) as freq from task where projectid=$projectId group by customerid;");
+                $rows = pg_fetch_all($result);
+
+                foreach ($rows as $row) {
+                    // There are chances that null shows up, lets skip it
+                    if ($row['customerid'] != null) {
+                        $taskCustomerFreq[$projectId][$row['customerid']] = $row['freq'];
```
jaragunde: There's only one project id being taken care for every loop, so you don't need a two-dimensional array. It could be just `$taskCustomerFreq[$row['customerid']] = $row['freq'];`.

update/update-from-2.16-to-2.17.php
```diff
@@ -0,0 +1,167 @@
...
+                    }
+                }
+            }
+
+            //We sort by value to get the highest number of customer association
+            foreach ($taskCustomerFreq as $projectFreq) {
+                arsort($projectFreq);
+                $taskCustomerFreq[$projectId] = $projectFreq;
+            }
```
jaragunde: As a consequence of the previous comment, you don't need this for loop. You can just call `arsort($taskCustomerFreq);`.

jaragunde: Afterthought: I think you should have used [asort](http://php.net/manual/en/function.asort.php) instead of [arsort](http://php.net/manual/en/function.arsort.php). Check the examples in php.net to see the difference more clearly.

jaragunde: Forget my previous comment, `arsort` is the correct one. I misread the examples!

update/update-from-2.16-to-2.17.php
```diff
@@ -0,0 +1,163 @@
...
+                    // There are chances that null shows up, lets skip it
+                    if ($row['customerid'] != null) {
+                        $taskCustomerFreq[$row['customerid']] = $row['freq'];
+                    }
+                }
+            }
+
+            //We sort by value to get the highest number of customer association
+            arsort($taskCustomerFreq);
+            if (isset($taskCustomerFreq)) {
```
jaragunde: `isset($taskCustomerFreq)` will always be true because you have set `$taskCustomerFreq = array()` in L126. Did you mean `if(count($taskCustomerFreq) > 0)`?

model/dao/ProjectDAO/PostgreSQLProjectDAO.php
```diff
@@ -610,9 +630,6 @@ public function partialUpdate(ProjectVO $projectVO, $update) {
         if ($update['areaId'])
         $sql = $sql . "areaid=" . DBPostgres::checkNull($projectVO->getAreaId()) . ", ";
 
-        if ($update['customerId'])
-            $sql = $sql . "customerid=" . DBPostgres::checkNull($projectVO->getCustomerId()) . ", ";
-
```
jaragunde: Why did you remove this? The customer id field in the project may be updated, just like any of the other fields.

jaragunde:
Some changes in the commit `[#271] Show up customers along with projects in tasks screen` broke the data management -> projects screen. Please check it out!

jaragunde:
Regarding `[#271] Show up customers along with projects in tasks screen`, projects without a client assigned now don't appear in the tasks screen, so there is no way to add tasks to them.

tonythomas01:
Need work here to push in the customerId too the database, which is not happening. Probably would need taking in the `customerId` from the name from the `combobox`, injecting it into the `taskRecord.customerId` data.

tonythomas01:
> Need work here to push in the customerId too the database, which is not happening. Probably would need taking in the customerId from the name from the combobox, injecting it into the taskRecord.customerId data.

That works now. I am currently saving the `customerId` along with `customerName` to the projectRecord, and this would make it easy for later retrieval.

jaragunde:
> I am currently saving the customerId along with customerName to the projectRecord, and this would make it easy for later retrieval.

I'm sorry to throw some work away, but this is not necessary. We don't want the customerId field to be part of the task any more. Every project will be assigned to only one customer now, so it's not necessary to have an additional relation between task and customer.

I prefer the customerId field to stay in the database, so admins can make their own double-checks after they run the migration script, but we should remove it from the application model (VO, DAO, etc.) or at least, make sure it's never saved again. It should stay "frozen in time" when the application is upgraded.

update/update-from-2.16-to-2.17.php
```diff
@@ -0,0 +1,163 @@
...
+            }
+
+            //We sort by value to get the highest number of customer association
+            arsort($taskCustomerFreq);
+            if (count($taskCustomerFreq) > 0) {
+                $customerAssociated = key($taskCustomerFreq);
+            }
+        }
+        // There can be projects which do not have a customer associated with it
+        if ($customerAssociated) {
```
jaragunde: This `if` is actually unnecessary, because you are always assigning `$customerAssociated = $customerList[0];` on L124. Projects with no customers associated will not appear in the `requests` table because it has not null restrictions.

update/update-from-2.16-to-2.17.php
```diff
@@ -0,0 +1,163 @@
...
+            $update_query = "UPDATE project SET customerid= $customerAssociated where id= $projectId";
+            print($update_query . "\n");
```
jaragunde: This was fine for debug purposes, but I'd write less technical message for final users. Something like "Project $projectName will be assigned to customer $customerName".

jaragunde:
Regarding your patches related to #271 (the customer+project combo box), I've noticed you are listing all projects in the tasks screen. In that screen, only open projects should be listed. In `projectDetails.php`, both open and closed projects should be listed.

web/js/projectDetails.js
```diff
@@ -80,22 +33,46 @@ var projectComboBox = new Ext.form.ComboBox({
...
+            }
+            return fn ? this.filterBy(fn) : this.clearFilter();
+        },
+        proxy: new Ext.data.HttpProxy({url: 'services/getProjectsAndCustomersForLoginService.php', method: 'GET'}),
         reader:new Ext.data.XmlReader({record: 'project', id:'id' }, projectRecord),
         remoteSort: false,
     }),
     mode: 'local',
     valueField: 'id',
     typeAhead: true,
```
jaragunde: This is causing some trouble when you type the name of the customer, because it will replace what you typed with the project name of the first entry. It should work just fine when this value is false.

web/viewProjects.php
```diff
@@ -85,6 +87,28 @@ function areas(val){
...
 
+    var customersStore = new Ext.data.ArrayStore({
+        id: 0,
+        fields: ['id', 'name'],
+        sortInfo: { field: 'name', direction: 'ASC' },
+        data : [
+            <?php
+
+            foreach((array)$customers as $customer)
+                echo "[{$customer->getId()}, '{$customer->getName()}'],";
```
jaragunde: There will be problems if the customer name contains a `'` character. You will have to escape those.

tonythomas01:
Removed all occurrences (as I could find out), and now we have: 
`➜  phpreport git:(remove-client) ✗ grep -R "RequestsDAO" * `
`➜  phpreport git:(remove-client) ✗ grep -R "DAOFactory::getRequestsDAO" *`
`➜  phpreport git:(remove-client) ✗ `

all returning null.

update/update-from-2.16-to-2.17.php
```diff
@@ -0,0 +1,160 @@
...
+
+    $customerAssignedToProject = array();
+
+    while ($row = pg_fetch_array($result, null, PGSQL_ASSOC)) {
+        $customerAssignedToProject[$row['projectid']][] = $row['customerid'];
+    }
+
+    foreach ($customerAssignedToProject as $projectId => $customerList) {
+        $customerAssociated = $customerList[0];
+        if (count($customerList) > 1) {
```
jaragunde: It would be good to print a message for those projects that are assigned to several customers, so the log of the migration script is more useful. If you only see messages about `Project $projectId will be assigned to customer $customerId` you can't tell apart which projects are actually interesting, the ones that were assigned to several customers.

The message could just be like: `Project $id is assigned to customers $id1, $id2, etc."

tonythomas01: done

jaragunde:
It looks we went a bit too far deleting things :) . Check the user tasks report, it's not working now. There are two different errors:
* 404 on `getCustomerProjectsService.php`
* 500 on `getTasksFiltered.php`, caused by: `Call to undefined method TaskVO::getCustomerId() in .../web/services/getTasksFiltered.php on line 247`

jaragunde:
Also the user details report still contains a project/customer report that it's obviously not working now. It should be replaced with a report that lists projects in rows and would only have a "total hours" column.

jaragunde:
This was a long one, but I think it's ready! Thanks!

PR #289, opened 2016-08-26
[#247] Improve the 'fresh-empty-task' on a new day functionality 

### LINKED ISSUE #247: Open empty days with a new, empty task
Description: Now when you click a date for the first time, there are not tasks and you have to press the button New task. The feature consists of adding an empty task the first time you click on a specific date instead.

Rationale: This change improves the user experience reducing time to track time.

tonythomas01:
Also show the `Status: New empty task created for editing` after creating a new empty task! Fix #247

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -103,6 +103,11 @@ var unsavedChanges = false;
 /* Variable to store if all tasks has loaded completely for a day */
 var loaded = false;
 
+/* Variable that will hold a fresh created task on page load */
+var freshCreatedTaskRecord = false;
+
+/* Variable that will hold a fresh panel, once created on page load */
+var freshCreatedTaskPanel = false;
 /**
  * Checks if the tasks for the day has loaded completely.
  *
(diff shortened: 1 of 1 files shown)


jaragunde:
First of all, if you write a commit message with three parts separated by commas, it probably means they belong to different commits :) Commits should address one single issue and they don't need a minimum size, it's enough if they make sense by themselves (e.g. the commit adds a feature or fixes a bug).

Now, regarding the patch:
* Create new task only for today: I don't think this suggestion makes sense, "today" should not be different from any other date.
* Respect task blocking feature: looks good, but I can't test it because the new task only pops up today :)
* Add delay: I think the most popular solution is b), we should go for that one.

jaragunde:
Results look good now! Just another suggestion to write cleaner patches: in my previous review, I suggested some changes to be done in your first patch, but you did those changes in your second patch and didn't mention that in the commit message. If somebody reads that patch in the future, they won't know where those small changes, unrelated with the commit message, came from.

In my opinion, the proper way to fix it would have been to amend the first commit to remove the delay and the "new task only for today" restrictions, and change the commit message to reflect that. Alternatively, sending a second commit containing those changes would have been the second best choice.

I'm merging the PR anyway, just take this into account for the future!

PR #292, opened 2016-09-03
[#249] Project combo box should list only projects assigned to user

### LINKED ISSUE #249: Project combo to list only projects assigned to user
Description: Now the project combo loads all the active projects where there is not customer selected. This change will consists of loading only the projects the user is assigned to and, additionally, create a new special entry called _show all_. When someone selects this entry, then, all the projects are loaded into the combo.

Rationale: Sometimes to find the right project takes time as many projects you do not need are loaded.

tonythomas01:
Fix #249 

More work needed on allowing view-all projects, probably by a button or checkbox.

### DIFF
--- a/model/dao/ProjectDAO/PostgreSQLProjectDAO.php
+++ b/model/dao/ProjectDAO/PostgreSQLProjectDAO.php
@@ -467,7 +467,7 @@ class PostgreSQLProjectDAO extends ProjectDAO {
         $sql = "SELECT project.*, customer.name AS customer_name FROM
                 project LEFT JOIN customer
                 ON project.customerid=customer.id
-                WHERE $activeCondition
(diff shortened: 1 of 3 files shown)


jaragunde:
Thanks Tony! The functionality of the checkbox looks good but the interaction with the keyboard should be polished (e.g. move the focus back to the combo box when it's checked).

Still, I don't really like how it looks and I would like you to try to implement a custom combo with an extra option, as described in #249. If you see it gets too complex or even impossible, we can always go back and retake the implementation with the checkbox.

To make your work easier, create another branch starting on commit `b862042` and send the alternative implementation with a custom combo as a different PR. We will merge the better branch from the two and close the other one.

tonythomas01:
Will start with the newbranch now.

jaragunde:
We finally merged the implementation in #295, closing this one.

PR #294, opened 2016-09-06
[#293] Fix: Template tasks may delete new tasks

### LINKED ISSUE #293: Template tasks may delete new tasks
I've just found a problem in the implementation of #247, in particular this part:

> b) delete the empty task when copying tasks from other days or creating the first task from a template.

Steps to reproduce:
1. Open a new, empty day.
2. Fill the empty task with some init/end time, project, etc.
3. Now click on a template button to add a second task.

Expected result: the template task is added, so there are two tasks now.
Actual result: the template task deletes the existing task, even when it's not empty any more.

tonythomas01:
Fix #293

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -1003,7 +1003,7 @@ Ext.onReady(function(){
                         dateString += cloneDate.getDate();
 
                         // If a fresh empty task exist, just remove it
-                        if(freshCreatedTaskRecord && freshCreatedTaskRecord.dirty) {
+                        if(freshCreatedTaskRecord && freshCreatedTaskRecord.dirty && isUnToched(freshCreatedTaskRecord)) {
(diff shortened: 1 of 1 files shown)


web/js/tasks.js
```diff
@@ -1318,6 +1318,13 @@ Ext.onReady(function(){
 
     summaryStore.load();
 
+    function isUnToched(taskRecord) {
+        if(taskRecord.get('text') == '' && taskRecord.get('descripton') == '' && taskRecord.get('initTime') == ''
```
jaragunde: `description` is not a field of `taskRecord`. Besides, there was a typo :D

I think it's enough with `text`, `initTime` and `projectId`.

jaragunde:
Now, the empty task never disappears when you create a task from template or use the "copy tasks" button.

Issue #144, opened 2016-03-09
In beta.phpreport, cloning a task and modifying the client, having a project inexistent for the client leads to the task being unmodifiable.


jaragunde:
[Trac import]
Reported by: agomez
Original date: Tuesday, 29 March 2011 10:42

1. Create a task with client and project.
2. Save.
3. Clone the just created task.
4. Modify the client in the cloned task to a client not having current selected project.

EXPECTED OUTCOME
================

I can keep modifying the rest of fields.

ACTUAL OUTCOME
==============

The complete task becomes dimmed and I'm not able to modify anything else. Not even delete the task!

Note: This is not happening with the stable version.

jaragunde:
[Trac import]
Comment by: jaragunde
Original date: Thursday, 05 May 2011 19:33

This bug can be reproduced skipping the step 3. Just:
 1. Create a task with client and project.
 2. Save.
 3. Change client.

The cause is that when changing the client, the project field remains unchanged. The case of the new client with the old project is detected like a task belonging to a closed project, and all fields are invalidated.

jaragunde:
[Trac import]
Comment by: jaragunde
Original date: Thursday, 05 May 2011 19:33

A workaround: deleting the project before changing the client.

jaragunde:
[Trac import]
Comment by: jaragunde
Original date: Thursday, 05 May 2011 20:29

```
commit 679d38ad49dbe9d21b423cc40770faaaee179487
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Thu May 5 20:20:03 2011 +0200

    [Bug #144] Fixed bug, verifying if the list of projects is empty because the user has changed the client.
```

PR #295, opened 2016-09-07
Smart custom combo with 'Load all Projects' item in dropdown

### LINKED ISSUE #249: Project combo to list only projects assigned to user

tonythomas01:
Fix #249

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -293,16 +293,25 @@ var TaskPanel = Ext.extend(Ext.Panel, {
                     reader:new Ext.data.XmlReader({record: 'project', id:'id' }, projectRecord),
                     remoteSort: false,
                     listeners: {
-                        'load': function () {
+                        'load': function (store) {
+                            var dummyRecord = new projectRecord({
+                                id: -1,                  // some invalid id
+                                description: "Load all projects",
+                                customerName: ""
+                            });
+
+                            store.add(dummyRecord);
+                            store.commitChanges();
+
                             //the value of projectComboBox has to be set after loading the data on this store
                             if ((this.findExact("id", this.parent.taskRecord.data['projectId']) == -1) &&
                                     (this.parent.taskRecord.data['projectId'] > 0)) {
                                 //no project with that id was found in this store
-                                if(this.baseParams.customerChanged) {
-                                    //the project could not be found because the user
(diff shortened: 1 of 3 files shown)


jaragunde:
There must be something strange with the internal status of the store. In some situations, I inadvertently erased the contents of the project field. Try these steps:
1. Fill one task in, select some project.
2. Save.
3. Reload.
4. Click on the project combo box of the saved task, then click somewhere else.
5. The application will detect a change in the task and will auto-save it.
6. Reload again. The project field is empty now.

web/js/tasks.js
```diff
@@ -293,16 +293,25 @@ var TaskPanel = Ext.extend(Ext.Panel, {
...
+                            store.commitChanges();
+
                             //the value of projectComboBox has to be set after loading the data on this store
                             if ((this.findExact("id", this.parent.taskRecord.data['projectId']) == -1) &&
                                     (this.parent.taskRecord.data['projectId'] > 0)) {
                                 //no project with that id was found in this store
-                                if(this.baseParams.customerChanged) {
+                                if(this.baseParams.login) {
                                     //the project could not be found because the user
                                     //has just changed the client
```
jaragunde: These code comments are not relevant. They belonged to the commit 679d38a, which fixed #144, but that doesn't apply anymore and I think this part of the code became useless. I understand you are reusing some useless code to implement the new feature, but please update the comments to explain it properly.

web/js/tasks.js
```diff
@@ -348,9 +354,17 @@ var TaskPanel = Ext.extend(Ext.Panel, {
                 forceSelection: true,
                 displayField: 'description',
                 tpl: '<tpl for="."><div class="x-combo-list-item" > <tpl>{description} </tpl>' +
-                        '<tpl if="customerName">- {customerName}</tpl></div></tpl>',
+                        '<tpl if="customerName"  >- {customerName}</tpl></div></tpl>',
```
jaragunde: I guess this is not needed :-)

PR #296, opened 2016-09-09
Do not show close confirmation when its just the empty task on screen

tonythomas01:
> (04:49:22  IST) mrobinson: tthomas: I had a suggestion for the "empty task on empty day" feature.
(04:49:35  IST) mrobinson: tthomas: Essentially make it inactive until you click on it.
(04:49:53  IST) mrobinson: tthomas: That way you won't get a lot of dialog boxes when you are clicking through empty days.

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -115,11 +115,27 @@ var freshCreatedTaskPanel = false;
 function isLoaded() {
     return loaded;
 }
+
+/**
+ * Check if a task is modified
+ *
+ * @param taskRecord
+ * @returns {boolean}
+ */
+function isUnTouched(taskRecord) {
+    if(!taskRecord.get('text') && !taskRecord.get('initTime') && !taskRecord.get('projectId')) {
+        return true;
+    }
+}
+
 /**
  * Checks if there are unsaved changes in this page.
  */
 function isUnsaved() {
-    return unsavedChanges;
+    if(!freshCreatedTaskRecord || !isUnTouched(freshCreatedTaskRecord)) {
+        return unsavedChanges;
+    }
+    return false;
 }
 
 function updateTasksLength(taskPanel) {
@@ -890,7 +906,7 @@ Ext.onReady(function(){
             },
         });
 
-        if(freshCreatedTask) {
+        if(typeof(freshCreatedTask) === 'boolean') {
             freshCreatedTaskRecord = newTask;
             freshCreatedTaskPanel = taskPanel;
         }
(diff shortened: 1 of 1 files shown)


jaragunde:
I've detected an undesirable situation:
1. Open a day that contains some tasks.
2. Modify the text of some existing task
3. Press the "new" button

Expected result: Ask before leaving the page
Actual result: Confirmation popup doesn't appear

Why is that? Because `newTask` function always receives the `freshCreatedTask` parameter as true. That's because you are assigning that function as a handler of several buttons, and those handlers receive several parameters like the button itself. When you run `newTask` after pressing a button, the `freshCreatedTask` parameter receives the button, and that evaluates to true.

tonythomas01:
Fixed that one. nice catch!

pabelenda:
Tested the corner case detected by Jacobo. Solved so far.

web/js/tasks.js
```diff
@@ -890,7 +906,7 @@ Ext.onReady(function(){
             },
         });
 
-        if(freshCreatedTask) {
+        if(typeof(freshCreatedTask) === 'boolean') {
```
jaragunde: You are only checking if `freshCreatedTask` is a boolean, but not if it's true! You should do both checks, otherwise it won't work in case you call newTask(false) :)

jaragunde:
I found another corner case:
1. Open an empty date.
2. Create a task from template.
3. Add some text in the description.
4. Change date.

Expected result: Ask before leaving the page, because there are unsaved changes
Actual result: Confirmation popup doesn't appear

The problem is when you run `isUnsaved()`, I think the variable `freshCreatedTaskRecord` is still pointing to the old object that contained the empty task, which was removed from the store when you created the template.

jaragunde:
Let me suggest to implement `isUnsaved()` in a different way, using `myStore.getModifiedRecords()`, I think it will be easier to understand and has less variables that must be properly kept in sync:

* If `modifiedRecords.length == 0` -> there are no unsaved changes
* If `modifiedRecords.length == 1`:
  * Check if `modifiedRecords[0] == freshCreatedTaskRecord`.
    * If they aren't equal -> there are some unsaved changes
    * If they are equal, and `isUnTouched(freshCreatedTaskRecord)` -> there are no unsaved changes
* If `modifiedRecords.length > 1` -> there are some unsaved changes in any case

You will be able to remove the `unsavedChanges` variable, you also won't need to reset `freshCreatedTaskRecord` in certain cases.

Issue #252, opened 2016-07-04
Add previous/next day buttons at the bottom of the tasks


jaragunde:
Description: At present to change of day, you have to look in the calendar and select the date you want to move. The feature consists of adding these buttons at the bottom of each task. The buttons should be also navigable with the keyboard.

Rationale: To click the exact dates in the calendar to navigate to previous and next days is slightly uncomfortable. With this feature, this can be solved.

tonythomas01:
Rather than putting it at the bottom of every task, we can use http://docs.sencha.com/extjs/3.4.0/#!/api/Ext.Panel-cfg-tools to put it somewhere neat at the top of page, with tooltips too! Looks good ? 

![selection_193](https://cloud.githubusercontent.com/assets/4214481/18408413/c1efcd52-774c-11e6-9149-cdb2e3730fde.png)

jaragunde:
```
commit e03f9b68bc0705b84b6a239caedf5da4b4266497
Merge: 25f85ea d60527c
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Tue Oct 25 20:09:22 2016 +0200

    [#252] Add previous/next day buttons and keyboard shortcuts
    
    (Merge pull request #298 from tonythomas01/add-next-prev-buttons)

commit d60527c16a9da966a64aa56c1cc586124cbdb0b2
Author: Tony Thomas <01tonythomas@gmail.com>
Date:   Mon Oct 24 19:35:03 2016 +0200

    Remove all custom tabIndex markers

commit 25bb3caf3e5545deecdb6ece5d3e23c963bf3008
Author: Tony Thomas <01tonythomas@gmail.com>
Date:   Wed Sep 14 15:38:30 2016 +0530

    [#252] Add navigation buttons and kill navigation while inputs or textarea in focus

commit f3c5b53eee56359dce5bee3f858ef3352afbecfe
Author: Tony Thomas <01tonythomas@gmail.com>
Date:   Tue Sep 13 17:32:12 2016 +0530

    [#252] Allow navigation to prev/next on keyboard left/right clicks

Issue #115, opened 2016-03-09
Show current editing day in the layout over the tasks


jaragunde:
[Trac import]
Reported by: chema
Original date: Wednesday, 22 September 2010 15:14

If you change the calendar month you can't see the day that is being currenlty edited, because the only reference is the one in the calendar, current editing date should be showed in the page.

An idea could be put a tittle over the task lists.

jaragunde:
[Trac import]
Comment by: jaragunde
Original date: Thursday, 23 September 2010 12:52

Although now the calendar shows the month to which the selected day belongs (that was in ticket #105 and it's fixed), adding a title would be a good idea.

jaragunde:
[Trac import]
Comment by: jaragunde
Original date: Monday, 27 September 2010 12:28

This isn't a defect anymore, since now users can always see the selected date marked in blue on the side bar calendar.

dmtrek14:
Closing since this is noted as no longer being a defect

PR #298, opened 2016-09-10
[#252] Add previous/next day buttons in Title panel header

tonythomas01:
selection_193

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -127,6 +127,22 @@ function isUnTouched(taskRecord) {
     }
 }
 
+// Navigate to Next day
+function navigateToNextDay() {
+    currentDate = new Date(date);
+    nextDate = new Date();
+    nextDate.setTime(currentDate.getTime() + 86400000);
+    window.location = "tasks.php?date=" + nextDate.format('Y-m-d');
+}
+
+// Navigate to Prev day
+function navigateToPrevDay() {
+    currentDate = new Date(date);
+    previousDate = new Date();
+    previousDate.setTime(currentDate.getTime() - 86400000);
+    window.location = "tasks.php?date=" + previousDate.format('Y-m-d');
+}
+
 function updateTasksLength(taskPanel) {
     if (taskPanel.initTimeField.getRawValue()!='' && taskPanel.endTimeField.getRawValue()!='' && taskPanel.endTimeField.isValid() && taskPanel.initTimeField.isValid()) {
         // We write the task's length label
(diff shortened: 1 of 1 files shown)


pabelenda:
I've detected something:

1. Go to September 2nd
2. Move to the left until you reach August 31st
3. Move to the right

Expected result: To be on September 1st
Actual result: To be on October 2nd

4a. If you move again to the right you will be on September the 3rd.
4b. If you move again to the left, you will be on September 1st.

tonythomas01:
Probably it was due to `addDate` adding in date irrespective of time, etc. 

changed to `addTime`, adding in `86400000`, which would do the trick.

pabelenda:
It is working quite nice.

pabelenda:
I've tested this.

addTime seems to work properly on day changes.

The day change using the arrow keys works nice as well.

jaragunde:
I'm afraid the original wording in the issue was terrible! I don't think the buttons should be on every task, because changing the date is not an action belonging to the task, but to the page.

I was actually thinking on something like this:
screenshot from 2016-09-27 18-42-58

(The top bar with the date is actually coming from a different ticket, #115).
Would it be much work to remove the buttons from every task and add them to the bottom bar, like in the image above? Otherwise, I could merge the PR as it is; I'd prefer to merge it as it is rather than not having the feature :)

tonythomas01:
The layout was changed after this PR, and we push the `new-task` thing to the center ?

tonythomas01:
I would say making a panel above that original panel to show the date might seem a bit overkill. How about showing the date in something like `date: 28-09-2016`  on top of the left dock (user work summary?)

jaragunde:
> I would say making a panel above that original panel to show the date might seem a bit overkill. How about showing the date in something like date: 28-09-2016 on top of the left dock (user work summary?)

That sounds good :) . We could still place the prev/next date buttons in the existing bar below the tasks. Imagine the situation where you are filling several days in a row; you could fill your tasks for a certain date and then jump to the next day to continue filling tasks, all using tab navigation.

tonythomas01:
Adding an icong to button seems overkill, so I ended up here. How does it look ? 

selection_207

jaragunde:
Just noticed the commit `[#252] Do not navigate while inputs or textarea in focus`, it contains two things: the part explained in the commit message and the addition of the next/prev buttons to the bottom bar. You should either split it or, at least, mention both things in the commit message.

jaragunde:
Oh, this rebase mess seems to be easily fixable with another `git rebase -i master`, making sure your master is up to date. Remember to remove any commits that shouldn't be there, you should only leave the last three ones.

jaragunde:
> Adding an icong to button seems overkill, so I ended up here. How does it look ?

Just to leave written testimony :) It looks good, we could just remove the arrow buttons in every task and make sure tab navigation lets users reach the buttons on the bottom. Then it would be ready to merge!

PR #300, opened 2016-09-10
[#194] Add hours column to user tasks report

### LINKED ISSUE #194: Add hours column to user tasks report
[Trac import]
Reported by: jaragunde
Original date: Monday, 18 March 2013 10:02

A column named 'hours' would be convenient to complement the information in init and end time columns. Its value would just be (end time - init time ) / 60.0.

Consider hiding the init and end time columns in the standard view of the report once the hours column is added.

tonythomas01:
Fix #194

### DIFF
--- a/web/js/userTasksReport.js
+++ b/web/js/userTasksReport.js
@@ -409,6 +409,7 @@ Ext.onReady(function () {
         {name:'date'},
         {name:'initTime'},
         {name:'endTime'},
+        {name: 'hours'},
         {name:'story'},
         {name:'telework'},
         {name:'onsite'},
@@ -459,6 +460,10 @@ Ext.onReady(function () {
             sortable: true,
             dataIndex: 'endTime',
         },{
+            header: 'Hours',
+            sortable: true,
+            dataIndex: 'hours',
+        }, {
             header: "Project",
             sortable: true,
             dataIndex: 'projectId',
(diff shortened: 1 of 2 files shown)


web/js/userTasksReport.js
```diff
@@ -527,46 +532,50 @@ Ext.onReady(function () {
...
         columnModel.setColumnWidth(2, 55);
-        columnModel.setColumnWidth(3, 200);
-        columnModel.setColumnWidth(4, 120);
-        columnModel.setColumnWidth(8, 120);
-        columnModel.setColumnWidth(9, 435);
+        columnModel.setColumnWidth(3, 55);
+        columnModel.setColumnWidth(4, 200);
+        columnModel.setColumnWidth(5, 120);
+        columnModel.setColumnWidth(6, 120);
+        columnModel.setColumnWidth(7, 435);
```
jaragunde: Please double-check the column width values, there are some mistakes. For example, the description column used to be 435 pixels width, but now you assigned that value to the onsite column.

tonythomas01: done.

jaragunde:
In the standard view, "task type" column should be hidden, and "story" should be shown instead.

jaragunde:
> In the standard view, "task type" column should be hidden, and "story" should be shown instead.

This was actually a regression introduced in d16c4fa8986585400db49adc97860359dc88d43b

PR #302, opened 2016-09-15
[#250] Change layout of fields in the tasks screen

### LINKED ISSUE #250: Change layout of fields in the tasks screen
Description: The idea is to create a new layout re-ordering and hiding some fields to optimize the space at the screen. The fields that will be hidden are:
* Task type
* Story
* Task story
* Telework
* Onsite

The layout to implement would be:

```
------------------------------
| 8:00 | 9:00 | project name |
-----------------------------| 
|   big description field    |
|                            |
-----------------------------|
|     COLLAPSED fields       |
------------------------------
```
Rationale: The goal of the feature is to save space and see more information at a glance in the page.

tonythomas01:
Fix #250

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -258,7 +258,8 @@ var TaskPanel = Ext.extend(Ext.Panel, {
             }),
             length: new Ext.form.Label({
                 parent: this,
-                ref: '../length'
+                ref: '../length',
+                style: 'display:block; padding:5px 0 5px 2px'
             }),
             endTimeField: new Ext.form.TimeField({
                 parent: this,
@@ -377,6 +378,7 @@ var TaskPanel = Ext.extend(Ext.Panel, {
                 mode: 'local',
                 valueField: 'id',
                 triggerAction: 'all',
+                width: 250,
                 forceSelection: true,
                 displayField: 'description',
                 tpl: '<tpl for="."><div class="x-combo-list-item" > <tpl>{description} </tpl>' +
@@ -491,6 +493,7 @@ var TaskPanel = Ext.extend(Ext.Panel, {
             taskStoryComboBox: new Ext.form.ComboBox({
                 parent: this,
                 tabIndex: tab++,
+                width: 180,
                 store: new Ext.data.Store({
                     parent: this,
                     autoLoad: true,  //initial data are loaded in the application init
@@ -551,8 +554,8 @@ var TaskPanel = Ext.extend(Ext.Panel, {
             }),
             descriptionTextArea: new Ext.form.TextArea({
                 parent: this,
-                height: 195,
-                anchor: '100%',
+                height: 110,
+                style: "width: 91%",
                 tabIndex: tab++,
                 value: this.taskRecord.data['text'],
                 enableKeyEvents: true,
@@ -571,7 +574,7 @@ var TaskPanel = Ext.extend(Ext.Panel, {
             deleteButton: new Ext.Button({
                 parent: this,
                 text:'Delete',
-                width: 40,
+                width: 100,
                 tabIndex: tab++,
                 margins: "7px 0 0 5px",
                 handler: function() {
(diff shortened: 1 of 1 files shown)


tonythomas01:
This is something I got here, do take a look and tell me how it is. 

It saves us from implementing the expand/collapse logics. 

selection_195

pabelenda:
Don't know how hard/necessary the expand/collapse functionality is.

In my humble opinion, visually, this is a nice solution.

aperezdc:
@tonythomas01: The layout change of tasks seems like a good idea. I like it from your proposed layout that it uses less vertical space... and it could probably use even less. The main textarea with the description could as well have just 2 or 3 lines of text, and there wouldn't be empty vertical space in the right side (between the “Template” button and the “Telework” checkbox). Also, if the logic for collapsing tasks is going to be removed, I would also remove the header bar which only says “Task” and is completely uneeded — we already know the thing is a task. The following is a quick UI mockup done in GIMP based on your screenshot:

phpreport-task

pabelenda:
@aperezdc I think that when @tonythomas01 wrote _It saves us from implementing the expand/collapse logics._, he was talking about what a different thing.

According to this comment (https://github.com/Igalia/phpreport/issues/250#issuecomment-243749035) on the initial report, the expand/collapse is intended to show/hide fields on the right side of the task panel. I think this is something we can skip, but this is just my opinion.

Regarding the expand/collapse of the tasks itself, I use it a lot. What I do is to have all the tasks collapsed, but the one I running at the very moment. I do not do this just because I want to save vertical space, for me, it makes things clearer. Again, this is just me.

aperezdc:
@pabelenda @tonythomas01: Ah, sorry for my mixing the different “collapse thing” features. At any rate, I still think it is worth it to reduce the vertical size of each tasks, even if task collapsing is kept.

tonythomas01:
Alright - so here is how it looks as of now - neat, less vertical space, and 7 row height straight with around 3 tasks showing up there on the single page. 

selection_200

jaragunde:
One of the points of this task was hiding the least used fields so it takes less tabs to fill the common fields (init, end, project and description, probably story too). Also, make the project entry bigger because our project names are quite wide now.

What about this?
99abecc0-7b59-11e6-8d28-197e04088094

Notice we are also hiding onsite and telework fields, which are not very important nowadays. The "more fields" button could be closer to the other buttons than in my example. Or we could forget about the collapsing logic and try this design with the bottom fields always visible first.

jaragunde:
Please, review tab navigation! It still follows the old sorting :)

jaragunde:
I have some thoughts regarding how it should adapt to screen widths. This is how it would look with a 1024x768 resolution, if you set your browser to use half of your full-hd screen, it would look similar:

screenshot from 2016-10-04 20-12-20

* The big text area should be able to adapt the width so the buttons are always on the right. Buttons could be smaller, since the text fits perfectly.
* I think the story combo should have a minimum width and the project combo should make room for it.
* Regarding spacing between items in the last row, I think we should let the padding grow with the width of the screen. Ideally, we should make the combo boxes grow too but it's not really important. We could leave this point for later because it doesn't impact functionality.

jaragunde:
BTW, this is how it looks when you use all the width of a full-hd screen:

screenshot from 2016-10-04 20-22-31

* It leaves quite a bit of unused space on the right of the buttons. The text area should probably grow bigger.
* At the same time, buttons would look better if there was some little space between them and the text area. There should be the same gap on their left and on their right.
* One more thought, items in the first row have some space between them but items in the last row don't. If they looked the same, the overall look would be much nicer, I think, even if you don't do anything else with the last row.

jaragunde:
> * Regarding spacing between items in the last row, I think we should let the padding grow with the width of the screen. Ideally, we should make the combo boxes grow too but it's not really important. We could leave this point for later because it doesn't impact functionality.

Actually, forget this comment now. Try to give the last row the same spacing between items there is between project and story in the first row. That should be easier, let's not make this too complicated yet O:-)

jaragunde:
Thanks for the patch, some comments:
* The template button is smaller than the other two, they all should have the same width.
* The layout is not good yet in a 1024px screen. Now buttons get partially hidden, and the end time combo jumps to the next line.

jaragunde:
A comment about code organization: in your commit `[#250] Show up large enough project and story input box on top`, the commit actually contains two things:
  * the size increase of the project and story input boxes
  * the creation of the bottom bar.

They should probably have been two different commits or, at least, that second part should have been mentioned in the commit message.

PR #305, opened 2016-09-16
[#297] Introduce getProjectsService and remove getProjectsAndCustomer…

### LINKED ISSUE #297: Unify the different getProject services
There are several services to retrieve data about projects in the system, with small differences in their API. These are the services I've found on a preliminary look:

* ~~getAllCustomProjectsService.php~~ - after #304 
* ~~getAllProjectsService.php~~ - renamed as `getProjectsService.php` after #305 
* ~~getFilteredCustomProjectsService.php~~ killed in #305
* getProjectService.php -- needs to be renamed to `getProjectDetails.php` 
* ~~getProjectsAndCustomersForLoginService.php~~ killed in #305 

The goals of this task are:
* Identify all the services that gather data from projects.
* Check the API of each of those services and design one service that covers all the use cases (or a couple of services, at most), removing the original ones.
* Change the frontend code to make use of the new service or services.
* The redesign will also probably propagate downwards in the application model. Check if it's possible to unify several facade or dao operations.

tonythomas01:
Fix #297


tonythomas01:
should've been two commits though :(

jaragunde:
I've rebased the commits and fixed some whitespace typos.

jaragunde:
After some testing, I detected the new services are ignoring the activation flag. For example, in the tasks screen the project combo box shows every project the user has been assigned, regardless they are currently active or not. If you press "load all", it will load all the projects in the system.

tonythomas01:
oh god :\ I think I pushed in the wrong things here. Let me correct them

jaragunde:
Be careful, you overwrote the fixes I had done to your patches when you pushed your new patch. You should have deleted your local branch and pulled the one from the server, then worked on top of that.

I've fixed this now, but take it into account in the future. This is not a common situation, though.

jaragunde:
I found a new issue; the "customer" filter in the project evaluation report is not working now.

tonythomas01:
also to be removed `ProjectsFacade::GetFilteredCustomProjects`

PR #307, opened 2016-09-21
[#235] Scroll to logged in username in acc hours report

### LINKED ISSUE #235: Scroll to user name in acc hours report
[Trac import]
Reported by: jaragunde
Original date: Thursday, 13 August 2015 11:29

In most cases, this report is used to check the user's own accumulated hours. It would be more useful if the table automatically scrolled to the name of the logged user and selected it to highlight the row.

As a further enhancement, the table should scroll to the same row that was selected after changing report dates. Even in case you are checking somebody else's hours, a reload would still show the hours for the same user.

tonythomas01:
Fix #235

### DIFF
--- a/web/viewWorkingHoursResultsReport.php
+++ b/web/viewWorkingHoursResultsReport.php
@@ -27,9 +27,10 @@
     define('PAGE_TITLE', "PhpReport - Working Hours Results Report");
     include_once("include/header.php");
     include_once("include/sidebar.php");
-
-
 ?>
+<script>
+var loggedInUser = '<?php echo $_SESSION['user']->getLogin(); ?>';
+</script>
 <script src="js/include/sessionTracker.js"></script>
 <script type="text/javascript" src="js/include/DateIntervalForm.js"></script>
 <script type="text/javascript" src="js/include/ExportableGridPanel.js"></script>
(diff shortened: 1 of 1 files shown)


jaragunde:
Unfortunately, this only works when the grid is reloaded. On the first load, no row is selected.

jaragunde:
More info: you need to use the viewReady event listener of the grid panel to select the default row the first time. The render event does not work for this purpose.

web/viewWorkingHoursResultsReport.php
```diff
@@ -145,6 +146,10 @@ function populate() {
                     var index = store.findExact('login', selectedLogin);
                     grid.getSelectionModel().selectRow(index);
                     grid.getView().focusRow(index);
+                } else {
+                    loggedInUserIndex = store.findExact('login', loggedInUser);
+                    grid.getSelectionModel().selectRow(loggedInUserIndex);
+                    grid.getView().focusRow(loggedInUserIndex);
```
jaragunde: Why create an else branch with almost the same code, when you could set `selectedLogin = loggedInUser` by default :)

jaragunde: If you do this, remember to remove the `selectedLogin = undefined;` assignment in the workingResultsForm view listener some lines below!

PR #321, opened 2016-10-14
Simplify query for tasks page with a filter flag after PR: #305

tonythomas01:
The patch fix various speed and user interaction issues after PR #305. Also fixed `projectDetails` report, which was not working after PR 305

### DIFF
--- a/model/dao/ProjectDAO/PostgreSQLProjectDAO.php
+++ b/model/dao/ProjectDAO/PostgreSQLProjectDAO.php
@@ -472,13 +472,14 @@ class PostgreSQLProjectDAO extends ProjectDAO {
      *        field.
      * @param string $filterCname string to filter projects by their customer name. NULL
      *        to deactivate filtyering by this field
+     * @param boolean $returnExtendedInfo flag to check if the response should include more information
      * @return array an array with value objects {@link ProjectVO} with their properties set to the values from the rows
      * and ordered ascendantly by their database internal identifier.
      * @throws {@link SQLQueryErrorException}
      */
     public function getAll($userLogin = NULL, $active = False, $orderField = 'id', $description = NULL,
-        $filterStartDate = NULL, $filterEndDate = NULL, $activation = NULL,
-        $areaId = NULL, $type = NULL, $filterCname = NULL) {
+        $filterStartDate = NULL, $filterEndDate = NULL, $activation = NULL, $areaId = NULL, $type = NULL,
+        $filterCname = NULL, $returnExtendedInfo = False) {
         $userCondition = "true";
         $activeCondition = "true";
         $conditions = "TRUE";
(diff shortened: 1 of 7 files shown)


model/dao/ProjectDAO/PostgreSQLProjectDAO.php
```diff
@@ -462,8 +462,6 @@ public function getByCustomerUserLogin($customerId = NULL, $userLogin = NULL,
      * @param DateTime $filterEndDate end date of the time filter for projects.
      *        Projects will a start date sooner than this date will be returned.
      *        NULL to deactivate filtering by this field.
-     * @param boolean $activation filter projects by their activation field.
-     *        NULL to deactivate filtering by this field.
```
jaragunde: The removal of the redundant `$activation` parameter should have belonged to a different commit. Remember, different fixes belong to different commits :)

model/dao/ProjectDAO/PostgreSQLProjectDAO.php
```diff
@@ -472,13 +470,14 @@ public function getByCustomerUserLogin($customerId = NULL, $userLogin = NULL,
      *        field.
      * @param string $filterCname string to filter projects by their customer name. NULL
      *        to deactivate filtyering by this field
+     * @param boolean $filter flag to check if the request has filtering variables set
```
jaragunde: I don't really like this parameter because it introduces an artificial and unexplained distinction among filtering variables, some would work even when `$filter` is false and some wouldn't.

You can check if the filtering variables you are interested are enabled or not, checking if they are null, and modify the query accordingly. For example, you may add the customer table only if `$filterCname != NULL`. Conditions, in general, will only add to the query time if the user has specified them.

Anyway, we actually need one parameter, that could be called `$returnExtendedInfo`, which would return all the dedication and cost information that comes from the extended query that involved the task table:
* If it is true, you will have to join the `task` and `hour_cost_history` tables, and add the `GROUP BY` clause.
* If it is false, you will have a simpler query without those two joins. It would default to false.

model/dao/ProjectDAO/PostgreSQLProjectDAO.php
```diff
@@ -489,39 +488,31 @@ public function getAll($userLogin = NULL, $active = False, $orderField = 'id', $
...
-                " OR project._end is NULL)";
-        }
-        if ($filterEndDate != NULL) {
-            $conditions .= " AND (project.init <= ".
-                DBPostgres::formatDate($filterEndDate) .
-                " OR project.init is NULL)";
-        }
-        if ($activation != NULL) {
-            $conditions .= " AND project.activation = " . $activation;
-        }
```
tonythomas01: It seems to make sense to not touch `$active` at all too, though.

model/dao/ProjectDAO/PostgreSQLProjectDAO.php
tonythomas01: Alright. Changed it to $returnExtendedInfo

tonythomas01:
Pushed in to the same commit, as no new changes were made.

jaragunde:
I think I didn't explain myself properly with regard to the `$returnExtendedInfo` parameter :)

In first place, it shouldn't affect `$description`, `$filterStartDate`, `$filterEndDate`, `$activation`, `$areaId` or `$type` selectors. Those should work regardless `$returnExtendedInfo` is true or false, since they don't imply any table joins. To sum up, those parameters should remain exactly as they were before!

About `$filterCname`, if it is enabled (different from null) it should add a join to the query. You could add a variable called `$customerJoin` (which would be `""` by default) to append the join to the query in case it is necessary.

Finally, in case `$returnExtendedInfo` is true, you should build the big query with the two joins (and an optional one as explained above) and run it with `customExecute`, otherwise you would run the smaller one (with one optional join for customer too).

jaragunde:
> About `$filterCname`, if it is enabled (different from null) it should add a join to the query. You could add a variable called `$customerJoin` (which would be `""` by default) to append the join to the query in case it is necessary.

As discussed on chat, @tonythomas01 is right and this join operation must always be there, so the ProjectVO object can always return the customer name with it. That simplifies things! `$filterCname` should just be treated like `$description`, `$filterStartDate`, etc.

web/services/getProjectsService.php
```diff
@@ -85,26 +94,31 @@
...
             $string = $string . "<end format='Y-m-d'>{$project->getEnd()->format("Y-m-d")}</end>";
-        else $string = $string . "<end/>";
-
-        $string = $string . "<estHours>{$project->getEstHours()}</estHours><type>" . escape_string($project->getType()) . "</type><movedHours>{$project->getMovedHours()}</movedHours><schedType>" . escape_string($project->getSchedType()) . "</schedType></project>";
-
+        } else {
+            $string = $string . "<end/>";
+        }
+
+        if($filter) {
```
jaragunde: Still using the old $filter variable here.

web/services/getProjectsService.php
```diff
@@ -85,26 +94,31 @@
...
-        if (!is_null($project->getInit()))
+        if ( !is_null( $project->getInit())) {
             $string = $string . "<init format='Y-m-d'>{$project->getInit()->format("Y-m-d")}</init>";
-        else $string = $string . "<init/>";
+        } else {
+            $string = $string . "<init/>";
+        }
 
-        if (!is_null($project->getEnd()))
+        if ( !is_null( $project->getEnd() ) ) {
```
jaragunde: Watch out for undesired style changes (probably made by the editor). These add noise to the patch and make it harder to read.

Issue #335, opened 2016-11-26
Right align the delete,clone, template buttons


tonythomas01:
Right now, they are crowded on the left side, and there is nothing on the right.

How about something like this --> 

![selection_245](https://cloud.githubusercontent.com/assets/4214481/20639597/3d8b87b4-b3c8-11e6-886e-3d8a63790602.png)

jaragunde:
My feedback from users is they didn't like buttons on the right side because they are too far away to reach them with the mouse in a wide screen window. Mainly because most buttons are on the left area of the application.

Ideally, I think we should not waste space and let fields use as much space as possible. For the bottom line I tried but it's a bit difficult, so in the end I left the **task type/task story/telework/onsite** fields with fixed size.

This is also related with #324 and #325, these tickets imply reducing the number of fields. The new field **tags** could expand to use the entire width of the bottom line.

tonythomas01:
> Ideally, I think we should not waste space and let fields use as much space as possible. For the bottom line I tried but it's a bit difficult, so in the end I left the task type/task story/telework/onsite fields with fixed size.

I think I found a way to apply % size to those dropdowns. Do you want it like spread all over ?

jaragunde:
> I think I found a way to apply % size to those dropdowns. Do you want it like spread all over ?

I have doubts, these fields don't seem to need bigger boxes since their contents already fit in the width we have given to them.

What if we right-align the _task type/task story/telework/onsite_ entries?

tonythomas01:
> What if we right-align the task type/task story/telework/onsite entries?

It might still make people to go to the right ? I thought we were avoiding it ?

jaragunde:
> It might still make people to go to the right ? I thought we were avoiding it ?

You are right! I forgot the point for a moment :)

Given how things are, it's better to leave things as they are now. I find your patch nicer to the eye, but it's not what our users want now.

In general, we will have to think at some point how to get the most of wide screens in our UI.

PR #336, opened 2016-11-26
[#335] Right align the delete-clone-template buttons in the bottomBox

tonythomas01:
A better alignment, and you can see the output here: 

selection_245


jaragunde:
Closing without merging, for reasons explained in #335. Thanks anyway!

Issue #352, opened 2017-06-07
"Pending holiday hours" should only report hours inside the report interval


jaragunde:
In the accumulated hours report, there is a "Pending holiday hours", which reports the number of holidays a user has available in the year in course and subtracts the amount of used holiday hours.

The problem is, this calculation is always done year-wide, without taking into account the report dates. For example:

* User has 100h of holidays this year.
* User has added 40h of holidays in the week 1 of the year.
* User has added another 40h of holidays in the week 30 of the year.
* If I check the acc. hours report between weeks 1 and 10, the reported number of pending holiday hours would be 20.
  * This is unexpected, as the holidays from the week 30 are outside the interval; the expected figure would be 60.

psaavedra:
a first look over this bug, looks like we could fix it according with the description with this simple change in the UI:

``` diff
diff --git a/web/viewWorkingHoursResultsReport.php b/web/viewWorkingHoursResultsReport.php
index 1251ae6..0059d25 100644
--- a/web/viewWorkingHoursResultsReport.php
+++ b/web/viewWorkingHoursResultsReport.php
@@ -216,7 +216,7 @@ Ext.onReady(function(){
                     if ($sid!="")
                         echo "&sid=" . $sid;
 
-                ?>&init=' + init.getFullYear() + "-01-01&end=" + end.getFullYear() + "-12-31";
+                ?>&init=' + init.getFullYear() + "-01-01&end=" + end.getFullYear() + "-" + (end.getMonth() + 1) + "-" + end.getDate();
 
                 pendingHoliday.load();
 
```


**Explanation of the issue:**

_Basically, the UI ignores the init and end dates and uses the entire current year._

jaragunde:
@psaavedra thank you, but that change alone is not enough. I've checked that the backend service will only calculate the amount of available holidays up to the selected date. An example:
* users have 100 hours of holidays for the entire year, full-time (`YEARLY_HOLIDAY_HOURS` config is 100)
* user has an 8-hour contract starting on Jan, 1st and ending on Dec, 31st.
* user has used no holidays yet.
* check the number of pending holidays between Jan, 1st and on June, 30st.
  * expected result: 100h
  * actual result: 49.59h

Why? Because the backend service should have calculated the pending holiday hours until Dec, 31st; then calculated the number of used holidays until the report date, and finally subtracted them.

psaavedra:
[...]

PR #358, opened 2017-08-08
[#352] Pending holidays computes all the entire last journey of the e…

psaavedra:
In model/facade/action/GetPendingHolidayHoursAction.php, removed the condition
which fix the end date of a journey interval to the end date passed as
parameter to the action DAO.

In case of the journey ends before the end of the year,
the pending hours will be prorated.

As working days always consider the entire journey ignoring the end date
interval ($this->end). This allows a correct and coherent count of the pending
holydays acording with the employee assigned journey.

Finally, In the /web/viewWorkingHoursResultsReport.php (UI) the interval for
the PendingHours request uses the submited data in the form instead of the end
date of the year.

Related issue: #352

### DIFF
--- a/model/facade/action/GetPendingHolidayHoursAction.php
+++ b/model/facade/action/GetPendingHolidayHoursAction.php
@@ -137,26 +137,38 @@ class GetPendingHolidayHoursAction extends Action{
             {
 
                 // First of all, we clip the interval with the journey
-                $init = $journeyRow->getInitDate();
-                $end = $journeyRow->getEndDate();
+                $initJourney = $journeyRow->getInitDate();
+                $endJourney = $journeyRow->getEndDate();
 
-                if ($init<$this->init)
-                    $init = $this->init;
+                if ($initJourney < $this->init)
+                {
+                    $initYearDay = date_create($this->init->format("Y") . "-1-1");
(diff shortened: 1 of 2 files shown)


jaragunde:
This implementation doesn't take into account situations like this one:
* User has a contract starting on 1-jan-2017 and ending on 31-dec-2018.
* We want to check his pending holidays on some date in 2017.
* Actual result: we get the unused holidays until the end of his contract instead of until the end of 2017.

jaragunde:
I think the `GetPendingHolidayHours(initDate, endDate)` operation should be reimplemented in these terms:
* overwrite initDate with 1-jan of the same year.
* calculate endOfYearDate as 31-dec for the same year in endDate.
* calculate workable hours between initDate and endOfYearDate and the proportional holiday hours between those dates.
* calculate the number of used holidays between initDate and endDate.
* return the number of proportional holiday hours minus the number of used holidays.

jaragunde:
Can you squash the two commits and merge? The first commit in the branch takes a different approach and I don't think it's worth keeping it, it would look confusing in the future. Some explanation on the final commit text would be welcome.

psaavedra:
Sure

psaavedra:
already squashed in a single commit.

Issue #357, opened 2017-08-03
New "last date" column in accumulated hours report


jaragunde:
In this report, knowing the last date when users introduced tasks in the system is interesting because it allows a more fair comparison. For example, if some person fills PhpReport weekly every Friday, this person will always have 4 negative days if you check the report on Thursday.

There will be a new column in the report, called "last date". It will contain, for every user, the latest date that contains any tasks and is inside the query interval.

For example, if we have two users who have filled their phpreport like this (0=empty, 1=some tasks):
```
|       | Mon | Tue | Wed | Thu | Fri |
| user1 |  1  |  1  |  1  |  1  |  1  |
| user2 |  1  |  1  |  0  |  0  |  0  |
```
* Query from Monday to Friday: last date for _user1_ is Friday, last date for _user2_ is Tuesday.
* Query from Monday to Wednesday: last date for _user1_ is Wednesday, last date for _user2_ is Tuesday.

psaavedra:
working on this change for this: https://github.com/Igalia/phpreport/commit/b4ff4f5e11dbd417a2a9c7bc8db9428f703184e9

This modifies the XML provided by `getExtraHoursReportService.php` adding a new element: `lastTaskDate`: 

``` xml
<reports>
<global>
<totalHours>4</totalHours>
<workableHours>2720</workableHours>
<extraHours>-2716</extraHours>
<totalExtraHours>-2716</totalExtraHours>
<lastTaskDate>2017-08-07</lastTaskDate>
</global>
<individual>
<report login="user">
<totalHours>0</totalHours>
<workableHours>0</workableHours>
<extraHours>0</extraHours>
<totalExtraHours>0</totalExtraHours>
<lastTaskDate>1900-01-01</lastTaskDate>
</report>
<report login="admin">
<totalHours>4.0000000000000000</totalHours>
<workableHours>1360</workableHours>
<extraHours>-1356</extraHours>
<totalExtraHours>-1356</totalExtraHours>
<lastTaskDate>2017-08-07</lastTaskDate>
</report>
</individual>
</reports>
```

jaragunde:
```
commit f25ffea9124eddc29c3fa89eca0f5f46fc2336d4
Merge: 5ac5585 56f678b
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Tue Aug 15 12:35:48 2017 +0200

    [#357] Implement last date column in accumulated hours report
    
    (Merge pull request #360)

[...]

PR #360, opened 2017-08-08
I357 new last date column in accumulated hours report

psaavedra:
* Added Last task date column in viewWorkingHoursResultsReport.php
* getExtraHoursReportService.php includes the lastTaskDate element
* Code formatting in getExtraHoursReportService.php

Related: #357

### DIFF
--- a/model/dao/TaskDAO/PostgreSQLTaskDAO.php
+++ b/model/dao/TaskDAO/PostgreSQLTaskDAO.php
@@ -974,10 +974,11 @@ class PostgreSQLTaskDAO extends TaskDAO{
         return $affectedRows;
     }
 
-    public function getLastTaskDate($userId, DateTime $referenceDate) {
+    public function getLastTaskDate($userId, DateTime $date, $strictBeforeDate=True) {
+        $operator = ($strictBeforeDate ? "<" : "<=");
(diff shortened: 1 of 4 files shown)


psaavedra:
**pending action**: commit squashing.

jaragunde:
Thanks for your patch, I've tested it. The dates returned by the web service via XML seem correct, but there is a problem when that date is parsed in the JS code: when printed in the grid, all the values are "today".

Also, the time is not necessary, I would format the dates like `dd/mm/yyyy`, to match the same format as the date selectors on top of the page.

psaavedra:
Umm, I can't replicate this problem which you describe. In my local database the last_date column matches with the day of the latest task done during the dates interval. What am I missing?

jaragunde:
Try this: create one task on Aug, 9 (yesterday) and another one on Aug, 10 (today). Then, go to the acc. hours report and set end date to Aug, 10 (or leave it blank for today). The expected "last task date" for that user is the 10 but it will say 9.

psaavedra:
Ok. I see.

jaragunde:
This is working properly now, and the "copy from yesterday" feature, which also uses `getLastTaskDate` works normally too. Proceeding to merge!

Issue #351, opened 2017-06-07
Don't allow to save tasks without an assigned project


jaragunde:
Every task should be assigned to some project. We must change this in ~two~ three levels:
* Frontend: the web UI shouldn't send tasks for save before the project field is set.
* Backend: tasks without a project shouldn't be saved. This affects third-party apps using the web services.
* Database: forbid null entries in the projectid field of a task row. If there are already any saved tasks with empty project ids, those should be migrated somehow. My first idea is to create a new project called "empty" and add those tasks there; then, the project "empty" can be checked to know who had empty project tasks and notify those users to reassign tasks properly.

jaragunde:
Edited ticket description to add the "database" level.

psaavedra:
@jaragunde, I need to know which is the database schema upgrade proccess and how works.

psaavedra:
Related PR: https://github.com/Igalia/phpreport/pull/361

jaragunde:
```
commit c238386d35cd712633e279fe39a0480c8bd99b72
Merge: bb8ccc9 88b5f20
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Wed Aug 23 11:16:19 2017 +0200

    [#351] Don't allow to save tasks without an assigned project
    
    Added controls to prevent this to frontend, web service backend and
    database.
    
    (Merge pull request #361)

commit 88b5f209806d2397f35672ce03f4303d56c5eb7d
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Wed Aug 23 11:09:39 2017 +0200

    Remove trailing whitespace and redundant comment lines

commit b384a709076b29e94e1cb33b94ae7a12c3a4843d
Author: Pablo Saavedra <psaavedra@igalia.com>
Date:   Wed Aug 16 15:21:19 2017 +0000

    Removed a trailing whitespace in createTasksService.php

commit e0d30227aca4709b4df59580da0f5b0d3e439d12
Author: Pablo Saavedra <psaavedra@igalia.com>
Date:   Mon Aug 14 16:09:59 2017 +0000

    Added SQL changes in the db schema

commit 8212a317924ded62aaad2e7d6d36f83cbcfecb99
Author: Pablo Saavedra <psaavedra@igalia.com>
Date:   Fri Aug 11 15:29:18 2017 +0000

    validateTasks in tasks.js also consider the projectComboBox

commit db2be051c12401f60eec10cd2f84afdd54425240
[...]

PR #361, opened 2017-08-11
[#351] dont allow to save tasks without an assigned project

psaavedra:
Related: #351 

DONE:
  * Frontend: the web UI shouldn't send tasks for save before the project field is set.
  * Backend: tasks without a project shouldn't be saved. This affects third-party apps using the web services.
  * forbid null entries in the projectid field of a task row.
  * set a default value for the previous entered tasks during the upgrade
  * if, there is tasks with no project assigned -> create a default "empty" project and assing the task to this project

### DIFF
--- a/web/services/createTasksService.php
+++ b/web/services/createTasksService.php
@@ -214,12 +214,17 @@
 
                 $taskVO->setUserId($user->getId());
 
-                $createTasks[] = $taskVO;
+                if (is_null($taskVO->getProjectId()))
+                {
+                    $string = "<return service='createTasks'><success>false</success><error id='4'>projectId is not valid</error></return>";
+                    break;
+                }
 
+                $createTasks[] = $taskVO;
             }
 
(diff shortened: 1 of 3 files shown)


psaavedra:
Changes in the database schema:

* Added the 'empty' project if there is tasks without project assigned
* Set already created task without project assigned to the 'empty' project
* projectid column in task table can not be null

web/services/createTasksService.php
```diff
@@ -214,12 +214,17 @@
...
+                    $string = "<return service='createTasks'><success>false</success><error id='4'>projectId is not valid</error></return>";
+                    break;
+                }
 
+                $createTasks[] = $taskVO;
             }
 
         } while ($parser->read());
-
+
```
jaragunde: There is a trailing whitespace here :)

psaavedra:
Removed.

Issue #332, opened 2016-11-10
Add direct access to user details for any user


jaragunde:
The menu entry *Reports -> User details* takes users to their own user details page, but it would be useful for managers to have access to the details of any user.

We should add an entry that opens a user selector and allows choosing a user to be queried. It should be allowed only to the manager role and upper. User details pages should only be allowed to that role, a basic user should only be able to check their own details.

psaavedra:
* check the role level in the frontend
* check the role level in the backend
* rename **Login**  as **Username**
* **Username** is a fixed value if logged user is a regular user
* **Username** is a text field if logged user has admin or manager level
* **Username** is a auto-complete text field

jaragunde:
You can see a similar behavior implemented already in _Reports -> User tasks_.

jaragunde:
```
commit d12464a9c33a1f5e47de57970907d4518e18fc1b
Author: Pablo Saavedra <psaavedra@igalia.com>
Date:   Wed Aug 16 20:59:27 2017 +0000

    [#332] Add direct access to user details for any user
    
    * Replaced 'login' field by 'userLogin' combobox
    * Added 'userStore' as Ext.data.Store for 'userLogin'
    * 'userLogin' is filled with the user accounts
    * 'userLogin' is only enabled when the logged user has the right
      permissions
    * Adjusted the width of the form panel
    * getUserProjectCustomerReportJsonService.php receives the uid selected in
      the 'userLogin'
    * Removed the Groups field in web/viewUserDetails.php
```

PR #363, opened 2017-08-16
I332 Add direct access to user details for any user

psaavedra:
* Replaced `login` field by `userLogin` combobox
* Added 'userStore` as Ext.data.Store for `userLogin`
* `userLogin` is filled with the user accounts
* `userLogin` is only enabled when the logged user has the right permissions
* Adjusted the width of the form panel
* `getUserProjectCustomerReportJsonService.php` receives the uid selected in the `userLogin`

Related issue #332

### DIFF
--- a/web/viewUserDetails.php
+++ b/web/viewUserDetails.php
@@ -56,38 +56,72 @@
 
     Ext.onReady(function(){
 
+    //schema of the information about users
+    var userRecord = new Ext.data.Record.create([
+        {name: 'id', type: 'int'},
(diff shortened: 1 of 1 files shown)


jaragunde:
The list of "groups" in the report doesn't add much information and, more importantly, it doesn't change when you change the user, showing misleading content. It's better to remove it.

psaavedra:
I see. It was a pretty good observation. Thanks for the revision. I just updated this branch with this requested change.

jaragunde:
Merged, thanks!

Issue #318, opened 2016-10-11
Only list active projects in project details combo by default


jaragunde:
In reports -> project details, there is a combo box where you can select any project. When there are many project, this combo gets too messy.

We should list only open projects in that combo by default, which are the most likely ones to be checked. Then, we should add an entry in the combo to list any project, including those currently inactive.

This should be similar to #249.

jaragunde:
The combo seems to be listing only active projects now (probably a side effect from the changes in the web services), but we should definitely add a "load all" entry.

jaragunde:
Reviewed at #364:
```
commit fdbece795e76d097d7ce39ff68558a76849e7c93 (HEAD -> master, origin/master, origin/HEAD)
Author: Pablo Saavedra <psaavedra@igalia.com>
Date:   Fri Aug 25 11:11:25 2017 +0000

    [318] Only list active projects in project details combo by default
    
    * Defined the 'user' variable in the scope of the web/js/projectDetails.js JS
      code
    * Modified the default behavior of the 'Project' combobox in
      projectDetails.php to: show only the active projects for the logged user
    * Added 'Load all active projects' entry in the 'Project' combobox in
      projectDetails.php
    * Added 'Load all projects' entry in the 'Project' combobox in
      projectDetails.php
```

Issue #330, opened 2016-11-08
"Load all projects" should be always present in the project combo box


jaragunde:
The project combo box has a special entry "Load all projects" to load all the open projects. When you type some characters and search on the combo is enabled, this entry is removed and only those projects matching the search criteria will be listed.

It would be better if "Load all projects" was always the last entry in the list, even when there are some search terms. When it was pressed, it would load the entire list of open projects and would apply the same search terms to the full list.

dmtrek14:
We are in the process of rewriting the frontend for tasks. Any outstanding bugs and enhancements for the old interface will be taken into account as we implement the new design.

PR #364, opened 2017-08-25
I318 Only list active projects in project details combo by default

psaavedra:
* Defined the 'user' variable in the scope of the web/js/projectDetails.js JS
  code
* Modified the default behavior of the 'Project' combobox in
  projectDetails.php to: show only the active projects for the logged user
* Added 'Load all active projects' entry in the 'Project' combobox in
  projectDetails.php
* Added 'Load all projects' entry in the 'Project' combobox in
  projectDetails.php

Make references to the #318 issue.


jaragunde:
Looks good, thank you! I wonder if users will miss #330 when this change arrives, time will tell.

Issue #326, opened 2016-10-26
Redirect to original URL after successful login


jaragunde:
When you click on some phpreport url (for example, the url to some project details) and you are not logged in, you will be redirected to the login screen. Nonetheless, after you successfully log in you will be redirected by default to the tasks screen instead of the url you wanted to visit.

The original url should be stored in a GET variable, and the login screen should use it for redirection once the user has entered their credentials.

jaragunde:
Reviewed at #365:
```
commit 8e2ee7e3d32a432e206c691bda4baed53f438dc5
Author: Pablo Saavedra <psaavedra@igalia.com>
Date:   Sat Aug 26 07:06:41 2017 +0000

    [#326] Redirect to original URL after successful login
    
    * auth.php pass a the $_SERVER[REQUEST_URI] as the
      'login_success_url' in the login.php redirection
    * added set_location_header_on_login_success_url() function in
      login.php
    * login.php check if a 'login_success_url' parameter is passed in
      the GET or POST and use it as Location value for the redirection
      on the login success
```

jaragunde:
I've come up with a situation we hadn't considered before: when the session times out (due to #65), it should also redirect to the page that the user had already opened. Right now, it redirects to the login page and then to the tasks of today.

To reproduce the problem, reduce the ping frequency in `web/js/include/sessionTracker.js` to just 1000 milliseconds (do a ctrl+shift+r reload to remove the cached version), open two phpreport tabs and log out from one of them; after a while, the other tab will notice the session has been closed. Press 'yes' to login again and you will see the problem.

Also, if you refresh (F5) the tab that has timed out, it won't redirect properly.

jaragunde:
Reviewed at #366:
```
commit dfdc0b7c0d2bd986a357c6b7dfd71741644064df
Author: Pablo Saavedra <psaavedra@igalia.com>
Date:   Thu Aug 31 17:00:40 2017 +0000

    [#326] Redirect to original URL after successful login
    
    The relogin modal dialog reloads the page instead of perform a
    redirection to the login page. If there is not a valid session the
    user will be finally redirected to the login page but passing the
    current loaded URL as parameter to come back again after the login
    success.
```

PR #365, opened 2017-08-26
I326 Redirect to original URL after successful login

psaavedra:
* auth.php pass a the `$_SERVER[REQUEST_URI]` as the `login_success_url` in the login.php redirection
* added `set_location_header_on_login_success_url()` function in  `login.php`
* `login.php` check if a `login_success_url` parameter is passed in  the `GET` or `POST` and use it as `Location` value for the redirection  on the login success

Related to the #326 issue


jaragunde:
This was a very annoying problem, thank you!

PR #366, opened 2017-08-31
I326 Redirect to original URL after successful login

psaavedra:
The relogin modal dialog reloads the page instead of perform a redirection to the login page. If there is not a valid session the user will be finally redirected to the login page but passing the current loaded URL as parameter to come back again after the login success.

Refers to #326


jaragunde:
Great solution, really simple. Thanks!

Issue #367, opened 2017-08-31
Hide Coordination menu


psaavedra:
The coordination menu is not always needed. Some deployments don't use this feautes and this menu entry obfuscate the usage of the application.

I'd suggest a new setting in the config file which active/deactive this entry.

jaragunde:
Merged commit reviewed in PR#368.

```
commit d4b26ae9c2797ff0a95335d302ce0f1d78683d25 (HEAD -> master, origin/master, origin/HEAD)
Author: Pablo Saavedra <psaavedra@igalia.com>
Date:   Thu Aug 31 18:31:41 2017 +0000

    [#367] Hide Coordination menu and 'task story' fields
    
    * Added the MENU_COORDINATION config parameter
    * When MENU_COORDINATION is set to FALSE, the Coordination menu in
      the sidebar is hidden
    * This two fields are hidden if MENU_COORDINATION is set to FALSE:
      * The "task story" entry in the tasks screen
      * The "task story" entry in the user tasks report, and the
        corresponding column from the grid.
      * Indentation revised
```

PR #368, opened 2017-08-31
I367 Hide Coordination menu

psaavedra:
* Added the MENU_COORDINATION config parameter
* When MENU_COORDINATION is set to FALSE, the Coordination menu in
  the sidebar is hidden

Refers to #367

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -699,6 +699,43 @@ var TaskPanel = Ext.extend(Ext.Panel, {
         this.teleworkCheckBox.setValue((this.taskRecord.data['telework']=='true'));
         this.onsiteCheckBox.setValue((this.taskRecord.data['onsite']=='true'));
 
+        bottomBoxTaskItems = [
+                        new Ext.form.Label({text: 'Task type', style: 'display:block; padding:5px 2px 5px 2px'}),
+                        this.taskTypeComboBox
+        ]
+
+        if (menuCoordination == true ) {
+            bottomBoxTaskItems.push(
+               new Ext.form.Label({text: 'TaskStory', style: 'display:block; padding:5px 2px 5px 2px'}),
+              this.taskStoryComboBox
+            )
+        }
+
+        bottomBoxItems = [
+            new Ext.Container({
+                width: 186,
+                layout: 'hbox',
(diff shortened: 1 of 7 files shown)


jaragunde:
This is a very interesting approach, thanks for coding it. There are more things that we could hide when the coordination feature is disabled:
* The "task story" entry in the tasks screen.
* The "task story" entry in the user tasks report, and the corresponding column from the grid.

jaragunde:
Ah, I would change the commit/PR title, because you aren't modifying the users evaluation and project summary entries.

psaavedra:
Oops, please go ahead.  It seems a copy and paste matter.

web/js/tasks.js
```diff
@@ -699,6 +699,43 @@ var TaskPanel = Ext.extend(Ext.Panel, {
...
+                    width: 186,
+                    layout: 'hbox',
+                    items:[
+                        this.deleteButton,
+                        this.cloneButton,
+                        this.createTemplateButton,
+                    ]
+                }),
+                new Ext.Container({
+                    width: 120,
```
jaragunde: Please use 145, like in the original code. That figure was tested with several browsers.
When the default font of the browser is slightly wider, the "onsite" text falls to the line below like this:
screenshot from 2017-09-04 12-08-47

web/js/tasks.js
```diff
@@ -699,6 +699,43 @@ var TaskPanel = Ext.extend(Ext.Panel, {
...
+                    layout: 'hbox',
+                    items:[
+                        this.teleworkCheckBox,
+                        this.onsiteCheckBox
+                    ]
+                }),
+                new Ext.Container({
+                    width: 460,
```
jaragunde: We should probably use a different width here when coordination is disabled. With my usual browser and window size, it looks like this:
screenshot from 2017-09-04 12-08-47

web/js/tasks.js
jaragunde: Also, check the indentation of this piece of code, it's mixing 8 and 4 spaces.

psaavedra:
Reverted the *width* value modification. The issue in the visualization of the labels seems it went away too

PR #377, opened 2017-12-05
Lock tasks before date

### LINKED ISSUE #345: Block task edition before a certain date
We currently have the ability to block tasks older than a certain number of days, but sometimes is useful to have additional criteria to block tasks. In particular, it would be good to be able to set a date before which tasks would be blocked.

The result would be that two criteria to block tasks would be available, admins should be able to enable them independently, and in case both are enabled the most restrictive criterion would be applied.

jaragunde:
Fixes #345.

Adds new fields to the database, means to retrieve and save their values from the frontend and reimplement `ConfigDAO::isWriteAllowedForDate` to take into account both a date and a number of days to block tasks.

Feedback is appreciated, specially on how to name the related fields to prevent confusion.

### DIFF
--- a/model/dao/ConfigDAO/ConfigDAO.php
+++ b/model/dao/ConfigDAO/ConfigDAO.php
@@ -92,9 +92,14 @@ abstract class ConfigDAO extends BaseDAO {
      *        so tasks older than a certain number of days would be blocked.
      * @param int $numberOfDays Set the number of days in the past when tasks
      *        tasks cannot be altered.
+     * @param boolean $dateLimitEnabled Enable of disable a limit date for tasks,
+     *        so tasks before that date would be blocked.
+     * @param DateTime $date Tasks before this date would be blocked if
+     *        $dateLimitEnabled is set.
      * @return boolean returns wether changes were saved or not.
      */
-    public abstract function setTaskBlockConfiguration($dayLimitEnabled, $numberOfDays);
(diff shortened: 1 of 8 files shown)


psaavedra:
Could I suggest this patch:

``` diff
diff --git a/web/settings.php b/web/settings.php
index 29a1940..5c9f361 100644
--- a/web/settings.php
+++ b/web/settings.php
@@ -79,7 +79,7 @@ if(!is_null($config['date'])) {
     echo "var date = Date.parseDate('" . $config['date']->format('Y-m-d') . "', 'Y-m-d');";
 }
 else {
-    echo 'var date;';
+    echo 'var date = "";';
 }
 if (isset($saved)) {
     echo 'var saved = true; ';
```

**Motivation**:

A empty value in *Date* (ref: web/settings.php) causes a 500 error in phpreport:

```
Argument 1 passed to DBAdapter::formatDate() must be an instance of DateTime, boolean given, called in /var/www/phpreport/model/dao/ConfigDAO/PostgreSQLConfigDAO.php on line 163 and defined in /var/www/phpreport/util/DBAdapter.php on line 206, referer: https://phpreport.localhost/web/settings.php
```

jaragunde:
Pushed a modified version of the last commit, including the requested change.

psaavedra:
+1

PR #379, opened 2018-01-19
Reimplement top menu bar as a list of links

### LINKED ISSUE #323: Top menu bar should contain regular links
The application menu bar on the top currently contains elements that trigger a JavaScript callback. As a result, you cannot use _right click -> open link in new window_ and similar features on them. This feels wrong, as you don't expect links to behave this way.

Let's replace them with regular links with an `href` field pointing to their actual destination, so we can use the menu more naturally.

jaragunde:
Implements the top menu bar as a list of links instead of using the ExtJS menu widget.

The new menu doesn't imitate the existing one. Instead, it's bigger and has a more plain aspect. It modifies the header and footer of the application to fit better.

~~Finally, an application icon is added in the top-left corner.~~ (this will be managed in a different PR).

Fixes #323.


jaragunde:
Pushed a new version of the patches with slightly modified changes in icon alignment and size.

jaragunde:
Removed application icon from the PR. It will be managed in a separate request.

pabelenda:
I've been taking a look into this on mi local environment. I have to say that this is something I really missed in phpreport. It looks superb from my POV, the only thing I miss is something telling me that the drop down buttons are precisely this. Like an small arrow pointing down.

https://www.w3schools.com/bootstrap/bootstrap_dropdowns.asp

jaragunde:
Just pushed a couple new commits, one to re-add the arrow icons and another one to re-add the divider lines that were also missing (I took the idea from that bootstrap examples link).

Notice there's a FIXME in the divider lines CSS; suggestions are welcome.

jaragunde:
Updated the commit about the arrow icons, I detected a problem: menus didn't open when you clicked exactly on the arrow.

pabelenda:
Regarding the fixme on the dropdown content divider, I don't know what to say apart from, why not just remove the divider completely?

Anyway, it is looking great to me. Merge! Merge! Merge!

Issue #382, opened 2018-01-30
Session activity check should handle temporary network disconnections


jaragunde:
This is how session activity check currently works:

* The session activity check (`web/js/include/sessionTracker.js`) periodically runs a callback that pings a certain backend service (`web/services/checkSessionActive.php`) to keep the session alive, and to check if it was closed by the server.
* When the server returns error, or the ping cannot reach the server, it interprets that the session is closed and shows a pop-up dialog telling the user to login again.

In the event of a temporary network disconnection, it's possible that the session activity checker interprets a closed session and opens the pop-up. When the network comes back, if it wasn't away for too long, it's likely that the session is still active, but the dialog telling the user to login again is already visible and won't go away.

We could enhance it adding these two steps:

* The callback keeps running periodically even after the "login again" popup has been shown.
* If it eventually gets a 200 response from the server, it should hide the visible popup.

pabelenda:
Opened this PR: https://github.com/Igalia/phpreport/pull/383

jaragunde:
{{{
commit 25b589acc1c4543aadf940b4709d004ea9490c80
Author: Pablo Abelenda <pabelenda@igalia.com>
Date:   Thu Feb 1 13:14:54 2018 +0100

    [#382] Added success block to hide MessageBox in case it exists.

}}}

PR #383, opened 2018-02-01
Added success block to hide MessageBox in case it exists

pabelenda:
I've been testing this on my local deployment.

Adding the `success` action is hiding the MessageBox in case it exists.

### DIFF
--- a/web/js/include/sessionTracker.js
+++ b/web/js/include/sessionTracker.js
@@ -28,6 +28,9 @@ function checkIfSessionExists() {
                     window.location.reload();
                 }
             });
+        },
+        success: function (response) {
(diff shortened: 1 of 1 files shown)


jaragunde:
Thanks for your patch. It looks good to me, just one nit: please follow the existing convention for commit messages, it would be: `[#382] Added success block to hide MessageBox in case it exists.`

pabelenda:
Ouch, my bad. I'll keep it in mind for the future.

jaragunde:
> I'll keep it in mind for the future.

You are still in time to amend the commit message (`git commit --amend`), and update the pull request (`git push --force`) :)

PR #386, opened 2018-02-15
Add application icon

### LINKED ISSUE #381: Create application icon
It would be great to create an application icon, something that we could show in the header and use as favicon or home page icon.

I've found this service that would help with the multi-browser setup of favicons, ideally we'd like that to work in as many browsers as possible: https://realfavicongenerator.net/ (found at [stackoverflow](https://stackoverflow.com/a/19590415)).

jaragunde:
Create application icon and use it in the header, as a favicon, etc.

Fixes #381.


jaragunde:
I had been unable to test the favicon, application icon, etc. until now because the configuration generated by https://realfavicongenerator.net/ assumes they are in the website root (recommended for compatibility). After setting this up, things are looking great; check this Firefox screenshot that shows the favicon and the application icon in the new tab page, compared with the look of other websites:
screenshot from 2018-03-02 11-33-54

PR #387, opened 2018-02-21
Change color scheme in accumulated hours report

### LINKED ISSUE #356: Change color scheme in accumulated hours report
In this report, any figure above 0 is green and below 0 is red. It would be more interesting, though, if color was used to highlight the most extreme accumulated hours values. For example:
* accumulated hours < -50: red
* accumulated hours between -50 and 50: green
* accumulated hours > -50: red again

The trigger value (in the example above, 50) should be configurable.

It is also not very interesting to apply colors to workable and worked hours, which are always positive anyway; they could remain black.

For holiday hours, the current scheme (green when positive, red when negative) could work ok.

jaragunde:
Implements a color scheme for the accumulated hours report as described in the ticket #356.

About the patches, I have doubts about the configuration value. I implemented it as a value in `config.php`, but maybe it should be in the config table of the DB and have a UI to modify it?

Fixes #356.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -184,3 +184,9 @@ define('VACATIONS_PROJECT', 'Holidays');
  * @global boolean Active the Coordination menu in the sidebar
  */
 define('MENU_COORDINATION', TRUE);
+
+/**
+ * @name EXTRA_HOURS_WARNING_TRIGGER
+ * @global int Value that acts as a warning trigger for extra hours values.
+ */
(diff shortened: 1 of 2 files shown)


jaragunde:
Pushed new version, fixed merge conflicts.

jaragunde:
My proposal is to merge this feature as it is now. We can transform the `config.php` value into something stored in the DB later if we deem it necessary.

Issue #389, opened 2018-03-02
Add a button to fill a full holiday


pabelenda:
At #327 a template button was added to ease the creation of a task with the correct start and end times so it would fill an entire work journey for the selected user and date.

This was intended mainly to fill holidays, but seems like the usage is broader. For that reason it would be useful then to create a template button that creates a full day task assigned specifically to the VACATIONS_PROJECT defined in the configuration.

jaragunde:
Reviewed in #390.

```
commit 71d414634f69f2eacec1960f2812876b964c0302
Author: Pablo Abelenda <pabelenda@igalia.com>
Date:   Wed Mar 7 13:03:29 2018 +0100

    [#389] Add template button to fill a full holiday
    
    This template sets the init and end date so it fills the entire
    journey for the current date. It also sets the project to be the
    holiday project specified in config.php.
    
    It triggers a save operation immediately to speed up holiday
    creation.
    
    (Squash pull request #390 from Igalia/full-holiday-task-button)
```

PR #390, opened 2018-03-02
[#389] Add a button to fill a full holiday

pabelenda:
Button added. Tested on a local deployment.

It will probably require some revision of the configuration files of the specific deployments, should we warn this somehow?

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -1200,6 +1200,37 @@ Ext.onReady(function(){
                     // Put the focus on the project field
                     taskPanel.projectComboBox.focus();
                 }
+            }),
+        // Default button for full-holiday task
+            new Ext.Button({
+		id: 'fullHolidayTaskButton',
+                text: 'Full-Holiday task',
+                disabled: forbidden,
+                handler: function () {
+                    // Delay task creation until store gets loaded or it won't be properly added
+                    if (!isLoaded()) {
+                        window.setTimeout(Ext.getCmp('fullHolidayTaskButton').handler, 100);
+                        return;
+                    }
+
+                    removeFreshEmptyTask();
+
+                    // Create task and fill init and end times
+                    var newTask = new taskRecord();
+                    var init = new Date();
+                    init.setHours(0, 0, 0, 0);
+                    var end = new Date();
+                    end.setHours(0, currentJourney*60, 0, 0);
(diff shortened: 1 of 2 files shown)


jaragunde:
I'm finding trouble with this patch. The task record gets created in a strange way, and the `createTask` call contains wrong information for project id:
```
<projectId>ext-record-644</projectId>
```

jaragunde:
Also, check the indentation in `tasks.php`, the existing code is indented with spaces but you used tabs.

jaragunde:
Let's not put the focus in the project field after the task is created, it's confusing.

jaragunde:
Also, an enhancement: let's save the task immediately after it's created! The trigger for auto save is 10 seconds, so if you want to fill e.g. one or two weeks of holidays in a row, it can be a bit annoying to wait 10 seconds (or manually save) every time.

pabelenda:
* Checked indentation for `tasks.php`
* Avoided focus set. It will remain in the button
* Added save logic. Fired it after 100 ms. Doing it straight away was failing.

jaragunde:
Now, the UI is reporting an error immediately before saving the task. There can even be several error messages if you created and deleted some tasks before:
* Create a task and save it.
* Delete the task, don't save yet.
* Create a holiday task with the "full holiday-task" button.
You will see several error messages. If you now delete the holiday task and then press the "full holiday-task" button again, the number of errors increases.

jaragunde:
`window.setInterval` will create a permanent callback that runs every 100 ms. Every time you press the button, a new callback is created, and they all step into each other's toes. Even with 100ms, chances are that the save operation is being called once again before the response arrives back, that's why there is an error message followed by the save confirmation.

jaragunde:
To run the save code snippet only once after 100ms, you can use `setTimeout`: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout

pabelenda:
Indeed `setInterval` is much more convenient.

jaragunde:
I'm still getting a validation error on the first save attempt. If I delay the timeout to 500ms, it works correctly. I think the problem here is the validation code, which doesn't properly run when the widgets of the task box are not ready.

Let's move forward with a higher timeout. We may need to adjust it for real deployments, widgets like the projects combo box may take longer to setup due to higher server latency.

Issue #341, opened 2017-01-26
 Improve user work summary at task panel 


thejoshwolfe:
There's a box called User Work Summary that shows a "Week goal" among other things. How about putting some or all of the information from `viewWorkingHoursResultsReport.php` for you in that box? The goal would be greater awareness of acc hours and holiday hours.

psaavedra:
The infomation that could be shown  (additionally to the current one):

* **Accumulated (Total) extra hours**: from 1-1-CURRENT_YEAR until TODAY (using the same logic implemented in `web/services/getExtraHoursReportService.php`)
* **Extra hours**: from Monday to Sunday of the current WEEK (using the same logic implemented in `web/services/getExtraHoursReportService.php`)
* **Workable hours**: from Monday to Sunday of the current WEEK (using the same logic implemented in `web/services/getExtraHoursReportService.php`)
* **Worked hours**: from Monday to Sunday of the current WEEK (using the same logic implemented in `web/services/getExtraHoursReportService.php`)
* **Pending holidays hours**: from Monday to Sunday of the current WEEK (using the same logic implemented in `web/services/getPendingHolidayHoursService.php`)

Maybe adding these new elements as part of the information provided in the `/web/services/getPersonalSummaryByDateService.php` action.

jaragunde:
In the accumulated hours report, the field "total extra hours" means the total amount of extra hours since the beginning of the contract. I would not use this term for anything else, so I would rename this field:
> **Accumulated (Total) extra hours**: from 1-1-CURRENT_YEAR until TODAY (using the same logic implemented in `web/services/getExtraHoursReportService.php`)

We could call it "extra hours (year)". In general, I think it's a good idea to make the time period explicit in the field name: worked hours (week), extra hours (week), extra hours (year), etc.

jaragunde:
There's a lot of data we could show, and the space is limited. Let's start with a user work summary that contains the following info:
* Worked hours (today)
* Worked hours (week)
* Goal (week)
* Extra hours (year)
* Pending holiday hours (year)
* Extra hours (total)

We would be removing the "month" item, which is the least useful.

psaavedra:
* +1

jaragunde:
Reviewed at #396:
```
commit ee5ef2f2f4c84d0a10859ed9f6b78c24ce77c9bf
Merge: 3364174a 861fce6b
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
[...]

PR #396, opened 2018-04-24
Improve user work summary at task panel 

pabelenda:
This implements the agreement we met at issue #341

### DIFF
--- a/docs/user/tasks.rst
+++ b/docs/user/tasks.rst
@@ -144,19 +144,33 @@ User work summary
 
 The *User work summary* panel on the left contains the following information:
 
-* Number of hours worked in the selected day, week and month.
+* **Worked (today):** Number of hours worked today.
 
-* Week goal for the selected week. This value is calculated taking into account
-  the users' journey values and the amount of worked hours so far. By default,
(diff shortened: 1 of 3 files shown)


jaragunde:
I've found some issues:
* Decimal positions are not being trimmed, leading to bad-looking numbers like in the picture.
* Some data are presented in hours:minutes and other in hours with decimal positions. We should unify them. Let's implement hours:minutes and wait for user feedback.
* Finally, we had talked about renaming the fields to make them more explicit. See https://github.com/Igalia/phpreport/issues/341#issuecomment-369553360

screenshot from 2018-04-25 14-05-20

pabelenda:
I've uploaded new changes.

I totally missed the field names. I thought the text inside parentheses was more like a clarification. The text are a bit large now, so I've added a bit of a tweak on the online css introduced at `web/js/tasks.js`.

I've treated timing more or less as done in the fields above. I've tested some cases and seems to be doing well.

pabelenda:
* Added a pointer to the user documentation
* Reformated and added user documentation to cover new fields and be more readable

jaragunde:
Results look great! Let's merge and wait for user feedback.

PR #398, opened 2018-05-07
Small changes over the User Work Summary

pabelenda:
* Changed field sort

Extra hours together. Pending Holidays to the bottom.

* New calculations over pendingHolidays.

If it is greater than a `week*currentJourney` it will show it in `day hours:minutes format` to improve readability

### DIFF
--- a/web/services/getPersonalSummaryByDateService.php
+++ b/web/services/getPersonalSummaryByDateService.php
@@ -129,7 +129,27 @@
         }
         if ($minutes < 10)
             $minutes = "0" . $minutes;
-        $pendingHolidays = $hours . ":" . $minutes;
+
+        $currentJourney = 8;
+        $journeys = UsersFacade::GetUserJourneyHistoriesByIntervals($initDate, $date, $userVO->getId());
+        if(count($journeys)==1) {
+                $currentJourney = $journeys[0]->getJourney();
+        }
+
+        if ($hours > $currentJourney*7) {
+            $days = intval($hours/$currentJourney);
+            $hours = round($hours - ($days*$currentJourney),2);
+            if ($hours == $currentJourney) {
+                $hours = 0;
+                $days = $days + 1;
+            }
+            if ($hours < 10) {
+                $hours = "0" . $hours;
+            }
+        $pendingHolidays = intval($days) . " days " . intval($hours) . ":" . intval($minutes);
+        } else {
+            $pendingHolidays = intval($hours) . ":" . intval($minutes);
+        }
 
         $string = "<personalSummary login='" . $userVO->getLogin() . "' date='" . $date->format($dateFormat) . "'><hours><day>" . $day  . "</day><week>" . $week  . "</week><weekly_goal>" . $weekGoal . "</weekly_goal><extra_hours>" . $extraHours . "</extra_hours><pending_holidays>" . $pendingHolidays . "</pending_holidays><acc_extra_hours>" . $accExtraHours . "</acc_extra_hours></hours></personalSummary>";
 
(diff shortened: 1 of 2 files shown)


web/services/getPersonalSummaryByDateService.php
```diff
@@ -129,7 +129,27 @@
         }
         if ($minutes < 10)
             $minutes = "0" . $minutes;
-        $pendingHolidays = $hours . ":" . $minutes;
+
+        $currentJourney = 8;
+        $journeys = UsersFacade::GetUserJourneyHistoriesByIntervals($initDate, $date, $userVO->getId());
```
jaragunde: You don't need to query the facade for this data, it's already been fetched in another part of the code: https://github.com/Igalia/phpreport/blob/master/web/tasks.php#L91

web/services/getPersonalSummaryByDateService.php
```diff
@@ -129,7 +129,27 @@
...
+        if(count($journeys)==1) {
+                $currentJourney = $journeys[0]->getJourney();
+        }
+
+        if ($hours > $currentJourney*7) {
```
jaragunde: I'd use $currentJourney * 5 (that means 40 in an 8-hour journey).

web/services/getPersonalSummaryByDateService.php
```diff
@@ -129,7 +129,27 @@
...
+            $days = intval($hours/$currentJourney);
+            $hours = round($hours - ($days*$currentJourney),2);
+            if ($hours == $currentJourney) {
+                $hours = 0;
+                $days = $days + 1;
+            }
+            if ($hours < 10) {
+                $hours = "0" . $hours;
+            }
+        $pendingHolidays = intval($days) . " days " . intval($hours) . ":" . intval($minutes);
```
jaragunde: You haven't re-calculated `$minutes` and this will give you wrong data if journey data is float.

web/services/getPersonalSummaryByDateService.php
jaragunde: Forget this comment, my mind was in the wrong layer ;)

web/services/getPersonalSummaryByDateService.php
jaragunde: A practical example, with a journey of 7.2h and 21.6h of holidays, that's exactly three days of holidays, but your code says "2 days 6:36 h".

jaragunde: Also, the indentation of this line is incorrect.

jaragunde:
Finally, it would be great if you could apply the same transformation into days, hours and minutes to the extra hours values. A small refactoring should help!

web/services/getPersonalSummaryByDateService.php
pabelenda: Indeed, 5 work days, not 7, that would mean a whole week.

web/services/getPersonalSummaryByDateService.php
pabelenda: will take a look asap!

pabelenda:
I'll just make a function that will convert a float number in hours into `hours:minutes` or `days hours:minutes` if it is big enough.

pabelenda:
I believe last changes are making thing look and work better

jaragunde:
Calculations still don't look right. I have this example: pending holiday hours are 102.60 and journey is 7.2. According to my figures, the result is 14 days, 1 hour and 48 minutes, but the application says "14 d 01:36 h".

pabelenda:
Maths are getting us harder today.

Issue #334, opened 2016-11-11
Send email reminders for users to fill their tasks in


jaragunde:
We could have a service to send an email to users who haven't filled their tasks in for a certain amount of time, for example one week. The amount of time and the contents of the message should be configurable.

jaragunde:
Let's define this task a bit more:
* Triggers are expressed as "number of days since the last saved task by the user".
* Warnings will be emailed to users and, optionally, to a list of CC recipients.
* The contents of the email will be customizable.
* There will be two triggers to send emails with different degrees of severity.
* Every trigger level will be configurable independently; will have a specific text and list of recipients.

jaragunde:
Reviewed at PR #400:
```
commit 25e95671af37cf1a3a7ff97210833befce161c23 
Merge: 5c4e79a1 63670af7
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Tue Jun 5 16:28:33 2018 +0200

    [#334] Creation of a email fill in reminder web services
    
    * Service to send reminders to users who haven't filled their tasks
      for a while.
    * Service for managers to check who's not filling PhpReport lately.
    
    (Merge pull request #400 from Igalia/no-fill-email-alert)

commit 63670af766b506c5c308c094902e21b6678432cd 
Author: Pablo Abelenda <pabelenda@igalia.com>
Date:   Tue May 22 15:54:06 2018 +0200

    [#334] Several changes over the no-fill alert service
    
     * Splitted the summary and individual emails in two jobs
     * Summary is meant to include all the people beyond boundaries
     * Added new template file to craft the managers summary
     * Excluded first warned people from the managers summary
     * Individual alerts are meant to be sent just one time each
     * Added an easy way to exlude users from the checks/report

commit 976f1b114c49c8625bee5fb4656c4bcac4dcc5ee
Author: Pablo Abelenda <pabelenda@igalia.com>
Date:   Tue May 8 08:53:06 2018 +0200

    [#334] Initial version of the Email reminder service
    
     * Added configuration parameters
     * Added a couple of email templates
     * Initial version of the web service

```

PR #400, opened 2018-05-07
Creation of a email fill in reminder web service

pabelenda:
To implement what is stated in #334 

 * Added configuration parameters
 * Added a couple of email templates
 * Initial version of the web service

### DIFF
--- a/jobs/sendEmailAlertNoFill.php
+++ b/jobs/sendEmailAlertNoFill.php
@@ -1,0 +1,112 @@
+<?php
+/*
+ * Copyright (C) 2018 Igalia, S.L. <info@igalia.com>
+ *
+ * This file is part of PhpReport.
+ *
+ * PhpReport is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PhpReport is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PhpReport.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** sendEmailAlertNoFill web service.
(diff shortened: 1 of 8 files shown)


jaragunde:
It looks good, but I've realized that any user would be able to trigger this service by opening the corresponding url `http://path/to/phpreport/web/services/sendEmailAlertNoFill.php`.

What means do we have to prevent web users to abuse this script? Maybe we could put it in a specific directory, and protect the web server from opening it? I suggest a `/jobs` directory in the repository root.

pabelenda:
I've moved the file and added a new email type: the last reminder before closing the month.

jobs/sendEmailAlertNoFill.php
```diff
@@ -74,80 +72,43 @@ function sendEmail($email, $lastTaskDate, $from, $cc, $subject, $template) {
...
-                    sendEmail($email, $lastTaskDate, $from, $cc, $subject, $template);
-                } elseif ($difference >= $NO_FILL_DAYS_TRIGGER_WARNING) {
-                    array_push($warnedUsers, $login);
-                    $cc = $NO_FILL_CC_WARNING;
-                    $subject = $NO_FILL_SUBJECT_WARNING;
-                    $template = $NO_FILL_TEMPLATE_WARNING;
-                    sendEmail($email, $lastTaskDate, $from, $cc, $subject, $template);
+        if (in_array($user->getLogin(), $excludedUsers))
+            continue;
+        else {
```
jaragunde: A couple of nits here:
* If you use `continue` , you don't need the else branch. Or you could have written the condition as `!in_array` with no else branch.
* I'd have used curly braces to reduce the ambiguity.

jobs/sendEmailSummaryNoFill.php
```diff
@@ -0,0 +1,102 @@
...
+
+    $users = UsersFacade::GetAllUsers();
+    $excludedUsers = [];
+    $criticalWarnedUsers = [];
+    $lastWarnedUsers = [];
+
+    foreach ($users as $user) {
+        if (in_array($user->getLogin(), $excludedUsers))
+            continue;
+        else {
```
jaragunde: Same here about if, else, continue.

pabelenda:
I've gathered all the changes into one single commit.

I've also changed what you suggested. I think it is better looking now, right?

PR #409, opened 2018-06-01
Vertical space saving on the left panel

pabelenda:
We've been adding information to the left panel. This has been putting pressure to the vertical available space. With the changes on this PR we save some of this space by doing:

 * Decreased top and bottom padding at the User Work Summary
 * Rearranged new and save buttons into one row layout
 * Rearranged collapse and expand buttons into onw row layout

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -1097,7 +1097,7 @@ Ext.onReady(function(){
 
 
     // Summary Panel
-    var userWorkSummaryTitleStyle = 'font-weight: bold; width:190px;';
+    var userWorkSummaryTitleStyle = 'font-weight: bold; width:190px; padding: 0px;';
     var summaryPanel = new Ext.FormPanel({
         width: 204,
         labelWidth: 100,
@@ -1105,8 +1105,8 @@ Ext.onReady(function(){
         frame:true,
         title: 'User Work Summary',
         defaults: {
-            labelStyle: 'padding: 2px 0 0',
-            style: 'text-align: right;',
+            labelStyle: 'padding: 1px 0 0',
+            style: 'text-align: right; padding: 1px 0 0',
         },
         tools: [{
             id: 'help',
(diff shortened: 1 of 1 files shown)


web/js/tasks.js
```diff
@@ -1097,15 +1097,15 @@ Ext.onReady(function(){
...
     var summaryPanel = new Ext.FormPanel({
         width: 204,
         labelWidth: 100,
         renderTo: Ext.get('summarypanel'),
         frame:true,
         title: 'User Work Summary',
         defaults: {
-            labelStyle: 'padding: 2px 0 0',
+            labelStyle: 'padding: 1px 0 0',
             style: 'text-align: right;',
```
jaragunde: You have to force the same padding rule in the `style` (that's the style applied to the right column), otherwise it defaults to 2px and you aren't saving any space.

web/js/tasks.js
```diff
@@ -1163,32 +1163,35 @@ Ext.onReady(function(){
...
-                    for(var i=0; i<panels.getCount(); i++) {
-                        panels.get(i).collapse();
-                    }
-                    cookieProvider.set('tasksCollapsed', true);
-                }
-            })
-        ]
+        //defaults: {
+        //    width: '100%',
+        //},
```
jaragunde: You left some commented code here.

web/js/tasks.js
```diff
@@ -1163,32 +1163,35 @@ Ext.onReady(function(){
...
+        layout: 'hbox',
+        items: [collapseButton, expandButton],
```
jaragunde: There was no need to create vars for the two buttons, you could have left the items definition unchanged. But if you are doing it for consistency with the rest of the code, that's fine.

web/js/tasks.js
pabelenda: Ouch! Sorry!

web/js/tasks.js
pabelenda: Indeed, that's the main reason. For me it is easier to read this way.

web/js/tasks.js
pabelenda: I'm not following. This style is applying and saving some space. I'm actually tempted to set it to `0px 0 0`

pabelenda:
Changes applied.

PR #424, opened 2018-09-04
Address some PHP warnings in project details report

jaragunde:
As a highlight of these changes, we started using the `??` null coalesce operator which is new to PHP7. It shouldn't be a problem to bump the required PHP version nowadays.


jaragunde:
Update with some "undefined constant" fixes that I had left behind.

jaragunde:
This also removes some warnings produced by other pages of the application, because of shared code in the backend.

jaragunde:
Added the issue number in the commit message and fixed some typos.

PR #452, opened 2019-04-24
Integrate with external authentication

jaragunde:
Add two configuration parameters to integrate with external authentication services. They will:
* Bypass password check, relying on the authentication service. PhpReport doesn't need to store passwords in this case. Use with caution!
* Use a different HTTP header to check for the authenticated user name.

### DIFF
--- a/web/login.php
+++ b/web/login.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2009 Igalia, S.L. <info@igalia.com>
+ * Copyright (C) 2009-2019 Igalia, S.L. <info@igalia.com>
  *
  * This file is part of PhpReport.
  *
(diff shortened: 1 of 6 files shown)


web/login.php
```diff
@@ -34,15 +34,21 @@ function set_location_header_on_login_success_url(){
...
 
+/* First check if a custom authentication header was set. */
+$customHeader = ConfigurationParametersManager::getParameter('EXTERNAL_AUTHENTICATION_USER_HEADER');
+if (!empty($customHeader) && isset($_SERVER[$customHeader]))
+    if(LoginManager::login($_SERVER[$customHeader]))
+        set_location_header_on_login_success_url();
+    else
+        echo _("Incorrect login information");
+
 /* There are Http authentication data: we try to log in*/
```
jaragunde: If the custom header is configured, execution should end here. Otherwise, in case of error in the login, it will try to log with the `PHP_USER_AUTH` header, with password check disabled.

It's hard that this can be exploited; I think it would require a valid user to pass the external authentication, then some problem with that user must happen in PhpReport login like, for example, that it doesn't exist in DB. Finally, adding an extra header, a different user could be impersonated. Anyway, it doesn't hurt to be extra-careful.

Issue #435, opened 2019-01-02
New users should be excluded from unfilled task reminders


jaragunde:
With the conditions defined at #334, new users will always receive these reminders. It should be easy to make an exception for them, we just have to check their contract list (journey or city assignment history) and compare with the oldest start date of the contracts.

jaragunde:
A (hopefully simpler) way to do this: when the user's last task is earlier than the configured deadline (or doesn't exist), before sending a warning check if the user had a contract on that day (there is a facade operation for that IIRC). This would also work for users with discontinuous contract periods.

jaragunde:
When I started working on this, I noticed new users were already excluded, because the script is supposed to be run daily and it checks if the date of the last task date *equals* to any of the triggers. If users don't have any last date, it won't trigger.

We can still make the script smarter for new users with some improvements. Reported them as #455.

jaragunde:
Pablo detected this is still an issue in the `sendEmailSummaryNoFill` job, because this one uses greater than/less than comparisons. Reopening, but without the "important" tag.

jaragunde:
Reviewed in PR #456:
```
commit 7b70ffa8581c31c2d9c93d1e987657f1dd3c1d8a
Merge: 82193425 e967534b
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Mon Jul 1 12:39:48 2019 +0200

    [#435][#455] Take into account contract info in unfilled task reminders
    
    (Merge pull request #456 from Igalia/i455-smarter-reminders)

commit e967534b93885bae943cfcbcb614e5e5cf7c72af
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Mon May 6 12:36:59 2019 +0200

    [#435][#455] Take into account contract info in unfilled task summary.
    
    Add equivalent modifications to the previous commit, to the job that
    generates a summary of users who have received warnings. This change
    also fixes newly created users being mistakenly listed.

```

PR #456, opened 2019-05-06
Take into account contract info in unfilled task reminders

### LINKED ISSUE #455: Unfilled task reminders should use contract information
We can make the script smarter for certain users by taking into account contract information:

* In the case of new users, warn them if they have been contracted for longer than the trigger time and haven't filled any tasks.

* If some user is not contracted any more, prevent warning them.

jaragunde:
Check contract start date for users without any tasks in the system, and exclude users who aren't currently contracted.

Fixes #455.


jaragunde:
Made a small enhancement in one commit, and added a third commit for the other job (I had forgotten about it). This will also fix #435.

Issue #478, opened 2020-08-06
Last month button in date filter


jaragunde:
It would be useful a button for "last month", in the same vein as the "last week" button in the date filter.

It would act with regard to the current date, and set a period between the first and last day of the month previous to the current one. E.g. we are now in August, so "last month" button would set init date to be the 1st of July and end date the 31st of July.

jaragunde:
Reviewed in PR #479:
```
commit e1d03571905679e20dda02240db265f8d520c38d (origin/master, origin/HEAD, master)
Merge: ac76d45e 960307f8
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Thu Dec 10 12:10:11 2020 +0100

    [#477][#478] New buttons for date filter.
    
    Merge pull request #479 from Igalia/i477-date-filter-new-buttons

commit 960307f82d4f14e273a3db74351c761f811cd2ad (origin/i477-date-filter-new-buttons)
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Thu Nov 26 11:06:30 2020 +0100

    [#477][#478] Modify DateIntervalForm size to fit new buttons.

commit 1ca3d8d95960f97e5568d2bf5d15ec5c197c90d8
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Fri May 15 11:15:56 2020 +0200

    [#478] Add last month button to date filter.
```

PR #479, opened 2020-08-06
New buttons for date filter

### LINKED ISSUE #477: Previous and next week buttons in date filter
To browse between week data faster, the date filter could contain a previous/next week couple of buttons. They would work like this:

* Previous week: It would select the week before the "start date" that has been set in the filter.
* Next week: It would select the week after the "end date" that has been set in the filter.

If the date filters are empty, buttons would act with regard to the current date.

jaragunde:
Adds previous week, next week and last month buttons to the date filter.

It would need some work to fit the buttons on screen, at the moment the buttons that don't fit fall into an "overflow" submenu.

Fixes #477 and #478.

### DIFF
--- a/web/js/include/DateIntervalForm.js
+++ b/web/js/include/DateIntervalForm.js
@@ -155,14 +155,14 @@ Ext.ux.DateIntervalForm = Ext.extend(Ext.Panel, {
             //default visual configuration
             frame: true,
             header: false,
-            width: 350,
+            width: 530,
             items:[{
                 id: 'form',
                 layout: 'form',
                 bodyStyle: 'padding:5px 5px 0',
-                labelWidth: 75,
+                labelWidth: 105,
                 defaults: {
-                    width: 230
+                    width: 380
                 },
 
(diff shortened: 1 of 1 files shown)


web/js/include/DateIntervalForm.js
```diff
@@ -229,6 +230,55 @@ Ext.ux.DateIntervalForm = Ext.extend(Ext.Panel, {
                         this._fireViewEvent();
                     }
                 }, '-', {
+                    text: '< Previous week',
+                    xtype: 'button',
+                    scope: this, //scope inside the handler will be the Form object
+                    handler: function () {
+                        var pivot = this._getStartDateField().getValue();
+                        if (!pivot)
+                            return;
```
jaragunde, line 239: This doesn't patch this part of the specification of the prev/next week buttons: "If the date filters are empty, buttons would act with regard to the current date."

jaragunde:
Finally decided to just increase the size of the container to fit the new buttons.

Issue #431, opened 2018-12-01
Fix PHP warnings


jaragunde:
There is a number of PHP warnings/notices in the web server log. These are just some of them:
```
[11-Sep-2018 11:24:37 UTC] PHP Notice:  Undefined offset: 0 in /var/www/html/phpreport/model/dao/ExtraHourDAO/PostgreSQLExtraHourDAO.php on line 126
[11-Sep-2018 11:24:37 UTC] PHP Notice:  Undefined index:  in /var/www/html/phpreport/model/dao/TaskDAO/PostgreSQLTaskDAO.php on line 409
[11-Sep-2018 11:24:37 UTC] PHP Notice:  session_start(): A session had already been started - ignoring in /var/www/html/phpreport/util/LoginManager.php on line 149
[11-Sep-2018 11:24:38 UTC] PHP Notice:  Undefined index: sid in /var/www/html/phpreport/web/services/getProjectsService.php on line 34
```

They should checked one by one and addressed.

jaragunde:
A number of patches addressing this kind of issues was merged in PR #424:
```
commit e5e16885d3a3483d868738e9d3fcd98473ac394f
Merge: f6a48312 2dc4214c
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Mon Dec 17 09:09:08 2018 +0100

    [#431] Address some PHP warnings in project details report
    
    (Merge pull request #424 from Igalia/php-warnings-project-details)

commit 2dc4214cc68cfe3f617082c726942d62d82837d6 (origin/php-warnings-project-details, php-warnings-project-details)
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Tue Sep 4 16:54:43 2018 +0200

    [#431] Fix some 'undefined index' warnings in PostgreSQLTaskDAO.
    
    Using is_null on an array element evaluates if the contents of that
    position in the array is null, assuming the element exists, and emits
    a warning if it doesn't.
    
    Using isset checks the existence of that element, which is what we
    actually want.
    
    Fixes warnings like this, logged from many phpreport pages:
    PHP Notice:  Undefined index:  in
    /var/www/html/phpreport/model/dao/TaskDAO/PostgreSQLTaskDAO.php on line 409

commit 7817586f8ecc13c1650c7d4f89f1ec6bcdd76b0a
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
[...]

PR #502, opened 2021-07-27
Fix a few PHP warnings

anarute:
Most of the issues were related to accessing an index from an empty array or null variable.

Tested locally with PHP 7.4.21

Related to #431

### DIFF
--- a/model/facade/action/ExtraHoursReportAction.php
+++ b/model/facade/action/ExtraHoursReportAction.php
@@ -185,7 +185,7 @@ class ExtraHoursReportAction extends Action {
                 $cityHistory = $cityHistoryDao->
                         getByIntervals($init, $end, $userVO->getId());
 
-                if (is_null($hoursWorked[0]['add_hours']))
+                if (!$hoursWorked || is_null($hoursWorked[0]['add_hours']))
                     $userWork[$userVO->getLogin()]["total_hours"] = 0;
                 else
                     $userWork[$userVO->getLogin()]["total_hours"] =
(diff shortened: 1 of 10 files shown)


jaragunde:
It looks great, thank you! Just one nit: the commit message in the second patch seems to be truncated... And, if you are going to reword past commit messages, maybe you could add "[#431]" to the first commit too! :innocent:

anarute:
> It looks great, thank you! Just one nit: the commit message in the second patch seems to be truncated... And, if you are going to reword past commit messages, maybe you could add "[#431]" to the first commit too! 😇

@jaragunde thank you for the quick review :) commits fixed!

Btw I had to rename the INSTALL file because I was in a very weird loop: I restored it and git deleted the install folder, if I restore the folder it deleted the file 😅  probably because they had the same name.

I had to add this change to this PR because I wasn't being able to rebase due to this (https://github.com/Igalia/phpreport/pull/502/commits/f7bafec45611b34d6477ba33335bab82c90b274e) but let me know if you prefer me to create a PR for it

jaragunde:
> Btw I had to rename the INSTALL file because I was in a very weird loop: I restored it and git deleted the install folder, if I restore the folder it deleted the file sweat_smile probably because they had the same name.
> 

Oh, maybe a file system limitation. Yes, it's a good idea to rename and have different names for the file and the folder, they could cause trouble somewhere else (Windows, too?).

anarute:
> > Btw I had to rename the INSTALL file because I was in a very weird loop: I restored it and git deleted the install folder, if I restore the folder it deleted the file sweat_smile probably because they had the same name.
> 
> Oh, maybe a file system limitation. Yes, it's a good idea to rename and have different names for the file and the folder, they could cause trouble somewhere else (Windows, too?).

And that's how I just found out that MacOS file system is case-insensitive by default 😬 

Btw added a new commit to fix a warning I just found

anarute:
FYI added a few more fixes since the last you've reviewed. I don't  intend to change this branch anymore, I'll create new PRs if I find new issues

jaragunde (approved):
Looks good, thanks!

PR #504, opened 2021-08-03
Remove customerid leftovers from TemplateVO

anarute:
It is not been used anymore and the property was already removed


jaragunde (approved):
Looks good to me

PR #505, opened 2021-08-04
Save init and end times in the task template

### LINKED ISSUE #346: templates should remember start and end time if set
The current behaviour of not setting them would be kept in case the fields aren't filled.

anarute:
Start storing init and end times to the task template.

If the task doesn't have times set, the template won't have them either.

Closes #346

### DIFF
--- a/model/dao/TemplateDAO/PostgreSQLTemplateDAO.php
+++ b/model/dao/TemplateDAO/PostgreSQLTemplateDAO.php
@@ -99,7 +99,7 @@ class PostgreSQLTemplateDAO extends TemplateDAO{
             throw new SQLIncorrectTypeException($templateId);
         $sql = "SELECT * FROM template WHERE id=".$templateId;
         $result = $this->execute($sql);
-        return $result[0];
+        return $result[0] ?? NULL;
     }
 
     /** Templates retriever by User id for PostgreSQL.
@@ -192,7 +192,7 @@ class PostgreSQLTemplateDAO extends TemplateDAO{
         }
 
         // Otherwise delete a task.
-        if(sizeof($currTaskVO) > 0) {
+        if($currTaskVO) {
             $sql = "DELETE FROM template WHERE id=".$currTaskVO->getId();
 
             $res = pg_query($this->connect, $sql);
--- a/model/dao/TemplateDAO/PostgreSQLTemplateDAO.php
+++ b/model/dao/TemplateDAO/PostgreSQLTemplateDAO.php
@@ -79,6 +79,8 @@ class PostgreSQLTemplateDAO extends TemplateDAO{
         $templateVO->setUserId($row['usrid']);
         $templateVO->setProjectId($row['projectid']);
         $templateVO->setTaskStoryId($row['task_storyid']);
+        $templateVO->setInitTime($row['init_time']);
+        $templateVO->setEndTime($row['end_time']);
 
         return $templateVO;
     }
@@ -130,7 +132,7 @@ class PostgreSQLTemplateDAO extends TemplateDAO{
     public function create(TemplateVO $templateVO) {
         $affectedRows = 0;
 
-        $sql = "INSERT INTO template (name, story, telework, onsite, text, ttype, usrid, projectid, task_storyid) VALUES(" .
+        $sql = "INSERT INTO template (name, story, telework, onsite, text, ttype, usrid, projectid, init_time, end_time, task_storyid) VALUES(" .
             DBPostgres::checkStringNull($templateVO->getName()) . ", " .
             DBPostgres::checkStringNull($templateVO->getStory()) . ", " .
             DBPostgres::boolToString($templateVO->isTelework()) . ", " .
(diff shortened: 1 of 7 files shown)


model/vo/TemplateVO.php
```diff
@@ -215,6 +219,30 @@ public function setTaskStoryId($taskStoryId) {
         $this->taskStoryId = $taskStoryId;
     }
 
+    public function getInitTime(): ?int {
```
anarute, line 222: @jaragunde Instead of documenting the function I thought of types notation for the parameters and return values like in this line, what do you think?
I find it more readable and it also makes the code more reliable - it will throw warnings and errors if we use pass wrong types

If we are using PHP 7.4 in the prod we can also type the properties making the code even cleaner: https://wiki.php.net/rfc/typed_properties_v2

jaragunde: I agree with using the types notation in new code. Thanks for the suggestion!

About typed properties, current minimum PHP version is 7.0, we will start requiring PHP 7.3 in the pull request #499. We could also bump it to 7.4, as it is widely available nowadays, but let's do that separately.

model/vo/TemplateVO.php
```diff
@@ -215,6 +219,30 @@ public function setTaskStoryId($taskStoryId) {
...
+    }
+
+    public function setEndTime(?int $endTime) {
+        $this->endTime = $endTime;
+    }
+
+    public function setEndTimeRaw(?string $endTime) {
+        $this->endTimeRaw = $endTime;
+    }
+
```
jaragunde, line 245: Where do you use the `set*TimeRaw` variations? I can see you use the getters to build the XML but I can't find where these are set.

web/services/createTemplatesService.php
```diff
@@ -27,8 +27,6 @@
 
 $request = trim(file_get_contents('php://input'));
 
-/*$request = '<?xml version="1.0" encoding="UTF-8"?><templates><template><customerId></customerId><projectId>1</projectId><ttype>community</ttype><story>wofff</story><taskStoryId></taskStoryId><telework>true</telework><onsite></onsite><text>wofff</text></template><template><customerId></customerId><projectId>1</projectId><ttype></ttype><story>asdf</story><taskStoryId></taskStoryId><telework>false</telework><onsite>true</onsite><text>woaowa</text></template><template><customerId></customerId><projectId>1</projectId><ttype></ttype><story>waowao</story><taskStoryId></taskStoryId><telework>true</telework><onsite></onsite><text>asdf</text></template></templates>';
-*/
```
jaragunde, line 31: I'd rather leave this line :innocent:: we have them in most service pages, and they serve as examples of the kind of XML that the service admits. Sorry it's a bit obscure; we should better document these services.

jaragunde (changes requested):
Thanks for the patches! I have some comments below.

web/services/createTemplatesService.php
anarute: Ok I won't fight for this one yet cause we don't have automated tests! hehe as soon as we do I would argue to remove all of these commented lines

anarute: commit removed

model/vo/TemplateVO.php
anarute: replied below https://github.com/Igalia/phpreport/pull/505#pullrequestreview-722419466

web/services/createTemplatesService.php
```diff
@@ -157,6 +155,47 @@
...
+                        if (is_null($initTimeFormat))
+                            $initTimeFormat = "H:i";
+                        $parser->read();
+                        if ($parser->hasValue)
+                        {
+                            $initTimeRaw = $parser->value;
+                            $initTimeParse = date_parse_from_format($initTimeFormat, $initTimeRaw);
+                            $initTime = $initTimeParse['hour']*60 + $initTimeParse['minute'];
+                            $templatesVO->setInitTime($initTime);
+                            $templatesVO->setInitTimeRaw($initTimeRaw);
```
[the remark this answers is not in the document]
jaragunde: Wow, how could I have missed that? Thanks!

model/dao/TemplateDAO/PostgreSQLTemplateDAO.php
```diff
@@ -188,7 +192,7 @@ public function delete(TemplateVO $templateVO) {
         }
 
         // Otherwise delete a task.
-        if(sizeof($currTaskVO) > 0) {
+        if($currTaskVO->getId()) {
```
jaragunde, line 195: This should be a null check, in case you try to delete a template that doesn't exist, line 191 will return null and line 195 will crash.

jaragunde (changes requested):
I detected one small issue:

model/dao/TemplateDAO/PostgreSQLTemplateDAO.php
anarute: Good catch, I just fixed this

jaragunde (approved):
Looks good to me!

PR #512, opened 2021-09-13
Add Holiday Management view

anarute:
In  the back end:
- create endpoints and methods to retrieve booked holidays for the logged user in a given period of time
- add a few automated tests to the method that groups the dates

In the front end:
- add Vue.js and V-calendar files
- add Vue component to display the booked holidays for the logged user in the current year

At the moment this view is read-only and will only display the booked holidays. Next step will be the creation and edition of the holidays in this same page.

Screen Shot 2021-09-13 at 15 27 31

This PR should be merged after #507 because of composer and autoload.


web/services/HolidayService.php
```diff
@@ -0,0 +1,106 @@
...
+{
+    private \LoginManager $loginManager;
+
+    public function __construct(
+        \LoginManager $loginManager
+    ) {
+        $this->loginManager = $loginManager;
+    }
+
+    public function datesToRanges(array $vacations): array
```
anarute, line 40: @jaragunde I tried to make this method as simple as possible but I know it's still a bit confusing, let me know if you can think of a better way to implement this :)

web/js/holidayManagement.js
```diff
@@ -0,0 +1,96 @@
...
+ * PhpReport is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PhpReport.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+var app = new Vue({
```
jaragunde, line 20: I will need to learn the basics of Vue so I can't really really review this now. But it seems simple enough, I like it!

web/js/holidayManagement.js
```diff
@@ -0,0 +1,96 @@
...
+                    },
+
+                },
+            },
+        };
+    },
+    created() {
+        const fetchData = async () => {
+            const currentYear = new Date().getFullYear();
+            const url = `/web/services/getHolidays.php?initi=${currentYear}-01-31&end=${currentYear}-12-31`;
```
jaragunde, line 56: I have my local dev environment at http://localhost/phpreport, and this component is not working because it's calling http://localhost/web/services/getHolidays.php

I had to modify this url to start with: `services/getHolidays.php...`

jaragunde (changes requested):
I had one problem with the holiday management screen, otherwise it's a great start!

I'll try to get things cleared out with the pending PRs ASAP, so you can continue.

web/js/holidayManagement.js
anarute: Oh I see.. just changed to `services/...` so we don't depend on the app setup

web/js/holidayManagement.js
anarute: Nice :) I really enjoy it, it feels simpler and more intuitive than React for me

web/services/getHolidays.php
```diff
@@ -0,0 +1,45 @@
...
+
+$holidayService = new HolidayService($loginManager);
+
+header('Content-Type: application/json; charset=utf-8');
+
+$holidays = $holidayService->getUserVacationsRanges($init, $end, $sid);
+
+if (array_key_exists('error', $holidays)) {
+    // user is logged out or doesn't have permission
+    http_response_code(403);
```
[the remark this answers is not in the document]
anarute: Yep! I was thinking of creating a REST API tiny structure, that would handle all the requests and route internally to each service based on the URL, but I'm still structuring it in my head.. probably will send a PR for this in the near future :)

model/dao/TaskDAO/PostgreSQLTaskDAO.php
```diff
@@ -655,6 +655,35 @@ public function getVacations(UserVO $userVO, DateTime $initDate = NULL, DateTime
 
     }
 
+    public function getVacationsDates(UserVO $userVO, int $projectId = NULL, DateTime $initDate = NULL, DateTime $endDate = NULL): array
```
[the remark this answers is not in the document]
anarute: Yeah, the idea of spiting was to use it in the other methods too, I thought of it when I started copy/pasting too much.

web/services/HolidayService.php
anarute: I've added a description with examples on top of it, let me know  if it's better

web/services/getHolidays.php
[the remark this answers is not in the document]
anarute: Expanding a bit further:

So instead of having the inumerous getXXXX, updateXXX  files that we have, we would have only the TemplateService, HolidayService, TaskService... and so on classes that would have all the methods for that "domain" (template, task, project, user...) and then a single file to handle the requests that would route based on the url and call the respective class service.

This request handler would take care of parsing the request, returning xml or json, checking authentication and permissions, anything related to the HTTP request. All the business rules should be inside the Service classes.

jaragunde (approved):

jaragunde (changes requested):
I found one small problem when trying this with real data: the list of holiday ranges on the left is filled with all the holidays ever used by that person, including previous years. That's not useful and, besides, it loads slowly when there are many.

anarute:
> I found one small problem when trying this with real data: the list of holiday ranges on the left is filled with all the holidays ever used by that person, including previous years. That's not useful and, besides, it loads slowly when there are many.

Good catch, just spotted a typo in the fetch url 😅

anarute:
> I found one small problem when trying this with real data: the list of holiday ranges on the left is filled with all the holidays ever used by that person, including previous years. That's not useful and, besides, it loads slowly when there are many.

@jaragunde this should be fixed now

model/dao/ProjectDAO/PostgreSQLProjectDAO.php
```diff
@@ -568,6 +568,18 @@ public function getAllCustom($active = False, $orderField = 'id') {
         return $this->customExecute($sql);
     }
 
+    public function getByDescription(string $desc): ?int
+    {
+        $sql = "SELECT * FROM project WHERE description='" . $desc . "'";
+
+        $res = pg_query($this->connect, $sql);
```
jayaddison, line 575: With the disclaimer that I worked with and developed a bit of an open-source crush on @anarute (sorry, Ana; I have to add this comment because I looked at this PR and can't in good conscience ignore it; I'm not following your day-to-day GitHub contributions, and this observation was a one-off)

It might be worth looking at [`pg_query_params`](https://www.php.net/manual/en/function.pg-query-params.php) to reduce the risk of SQL injection here; it should handle escaping of parameters to the query.

anarute:
@jaragunde I just pushed 2 new commits to skip the vue app from minify, can check if they solve the issue, please?

jaragunde:
It works, thanks. Any particular reason to move holidayManagement.js to the vuejs directory? It doesn't seem to be necessary to make the minification script work again.

anarute:
> It works, thanks. Any particular reason to move holidayManagement.js to the vuejs directory? It doesn't seem to be necessary to make the minification script work again.

@jaragunde when I was testing locally if I didn't move it the ugligy was trying to minify it and failing

jaragunde:
If I go back to the second-to-last commit it seems to work for me, `holidayManagement.min.js` is generated although it's never used because you are excluding `holidayManagement.php` from the files we run `sed` on.

I would prefer not to move it so we have all the code in the places we expect, but we can still make the exceptions you made for `holidayManagement.php` to keep things simple inside our possibilities. Could you please just remove the last commit?

We definitely must use more robust, modern tooling for this. So much to do... :slightly_smiling_face:

anarute:
> If I go back to the second-to-last commit it seems to work for me, `holidayManagement.min.js` is generated although it's never used because you are excluding `holidayManagement.php` from the files we run `sed` on.
> 
> I would prefer not to move it so we have all the code in the places we expect, but we can still make the exceptions you made for `holidayManagement.php` to keep things simple inside our possibilities. Could you please just remove the last commit?
> 
> We definitely must use more robust, modern tooling for this. So much to do... 🙂

Oh sorry, I completely missed this comment, just did  it.

jaragunde (approved):
Looks good to me now!

jaragunde:
Let's merge this PR so you can keep working on this feature more easily. Some things to take care of in the next patches:
* Show holiday hour amounts as used/planned/unplanned.
* List days that have *no full-time* holidays saved in, with the specific number of hours (e.g. "2021-09-24: 4 hours").
* If we are short of space in the left bar, we might remove the date ranges. I think they are not that useful if we have the ranges in the calendar and the total amounts.
* We might want to double-check that any DAO functions we modify our add (DAOs or anywhere, actually) add type restrictions to their parameters and return types to modernize the code base progressively.
* And, of course, the ability to *save* holidays from this screen, which will have its own problematic with part-times... But we can talk about that later, I think it will be easier to start with the most common scenario and progressively polish it to accommodate more complex cases.

PR #515, opened 2021-09-23
Add dateBelongsToJourney method to JourneyHistoryVO

anarute:
This method checks if a given date is between the init and end dates of a journey.

It will be used to find what journey a user had in the date of a specific task.

@jaragunde I thought of adding this "utility" methods next to their VO classes, let me know if you think they should be somewhere else.


jaragunde (approved):
Looks good! Just one comment, remember there is a parent class `BaseHistoryVO`, where the date attributes are stored. In case we ever need this feature in another `*HistoryVO`, let's not forget to move this method up instead of writing a copy.

Issue #510, opened 2021-09-13
Improved holiday management


jaragunde:
We would like to improve holiday management in the following ways:

* Create a UI where users can set multiple days of holidays from a single UI. This UI would show a year-long calendar (similar to calendar management page) and users would create holiday tasks just clicking on the corresponding days.
* We would add means to display "enjoyed" holidays, "scheduled" holidays and "unscheduled" holidays differently. Enjoyed holidays would be considered those assigned to days in the past, while scheduled holidays are those which are saved into a day in the future. Unscheduled holidays would be the ones remaining to fill the yearly total (yearly total - enjoyed - scheduled).
* We would create reports for managers to check the different types of holidays for certain groups of people. They will need to know the amount of enjoyed, scheduled and unscheduled  holidays for every user, and get a per-week planning.

jaragunde:
Merged commits from PR #512:
```
97e14caa [#510] Add Holiday Management view [Jacobo Aragunde Pérez]
1dfd5478 Remove holiday app from minify script [Ana Rute Mendes]
d11b0309 Fix typo in getHolidays url [Ana Rute Mendes]
0f9cd825 Add comments to datesToRanges to explain what it does [Ana Rute Mendes]
40f41a67 Update getHolidays url [Ana Rute Mendes]
6822bec3 Add HolidayManagement view [Ana Rute Mendes]
45ecd139 Add a few tests for datesToRanges [Ana Rute Mendes]
3352c066 Add Vue.js and v-calendar files [Ana Rute Mendes]
e44d36be Add getHolidays endpoint [Ana Rute Mendes]
c9ffb569 Add HolidayService [Ana Rute Mendes]
d191e447 Add GetScheduledHolidays action [Ana Rute Mendes]
c4849a3f Add GetProjectByDescription action [Ana Rute Mendes]

**Comment by jaragunde** [comment]
Merged commits from PR #518:
```
d9ecbc87 [#510] Implement holiday editing methods [Jacobo Aragunde Pérez]
c9b508ea Add button styles [Ana Rute Mendes]
c26d97cd Remove list of dates [Ana Rute Mendes]
f81ea048 Implement update vacations method [Ana Rute Mendes]
e1c14c90 Abstract method to update dates and ranges in the view [Ana Rute Mendes]
24a7f512 Add createVacations method [Ana Rute Mendes]
100f8c24 Add isWeekend method [Ana Rute Mendes]
[...]

PR #518, opened 2021-09-27
Implement holiday editing methods

anarute:
Use the same view that displays the holidays yearly to allow the user to add and edit existing holidays.
Given an array of days:
1 - delete the vacation days that are not in the list but were booked previously by the user.
2 - create new holiday tasks from the array that weren't booked yet.

Note: Holidays selected in weekends are ignored, holidays that don't belong to any journey are also ignored (for instance, trying to book a holiday in a date previous to when the person joined the company).

Next steps of the implementation:
- distinguish half leave from full day holidays visually
- display more information in the sidebar
- display success/failure messages after saving

Part of #510

Example:
- saving valid holidays and then trying to save holidays in a period before the user joined the company fails

https://user-images.githubusercontent.com/333447/134967700-0149d455-dfaf-4476-bbc0-46974282921b.mov

### DIFF
--- a/web/js/holidayManagement.js
+++ b/web/js/holidayManagement.js
@@ -17,6 +17,37 @@
  * along with PhpReport.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+function formatDate(date) {
+    // Receives a date object and returns a string in the format YYYY-MM-DD.
+    // The subtract below is to discard timezone.
+    return new Date(date.getTime() - (date.getTimezoneOffset() * 60000))
+        .toISOString()
+        .split("T")[0];
+}
+
+function deleteRangeAndDays(ranges, day, days) {
+    rangeIdx = ranges.findIndex(range => range.coveredDates.includes(day));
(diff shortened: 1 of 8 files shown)


web/js/holidayManagement.js
```diff
@@ -17,6 +17,37 @@
...
 
+function formatDate(date) {
+    // Receives a date object and returns a string in the format YYYY-MM-DD.
+    // The subtract below is to discard timezone.
+    return new Date(date.getTime() - (date.getTimezoneOffset() * 60000))
+        .toISOString()
+        .split("T")[0];
+}
+
+function deleteRangeAndDays(ranges, day, days) {
```
jaragunde, line 28: Sorry, I don't understand the purpose of the `days` parameter here. Could you please add some comments? A short line about the purpose of the function, its inputs and outputs would be great.

jaragunde (approved):
The comment below is just a suggestion, I'll approve and merge so I don't block you any longer. Thank you!

PR #519, opened 2021-09-29
[#431] Fix PHP warnings in user management screen.

jaragunde:
Fixes:
* 'undefined variable' errors in web services layer.
* 'sizeof(): parameter must be an array' errors in DAO layer.
* 'undefined offset' errors in DAO layer.

### DIFF
--- a/web/services/updateUserGoalsService.php
+++ b/web/services/updateUserGoalsService.php
@@ -40,7 +40,7 @@ do {
 
     $parser->read();
 
-    if ($parser->name == 'updateUserGoals')
+    if ($parser->name == 'userGoals')
     {
 
         $sid = $parser->getAttribute("sid");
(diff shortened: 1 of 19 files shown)


web/services/updateUserGoalsService.php
```diff
@@ -40,7 +40,7 @@
 
     $parser->read();
 
-    if ($parser->name == 'updateUserGoals')
+    if ($parser->name == 'userGoals')
```
anarute, line 43: Why did you have to change this one?

jaragunde: It was wrong, the top-level tag for this service is "userGoals" and for that reason it never initialized the `$sid` variable like other similar services do:
```
<?xml version="1.0" ?>
<userGoals sid='12341235123591'> <!-- sid is optional, actually never used -->
  <userGoal>
    <userId>2</userId>
    <id>2</id>
    <extraTime>11</extraTime>
    <init>2021-09-30</init>
    <end>2022-09-28</end>
  </userGoal>
</userGoals>

anarute (approved):

Issue #432, opened 2018-12-01
NaN in weekly goal 


jaragunde:
Open phpreport in 31st Dec, 2018; the displayed weekly goal is "NaN".

I suspect the problem is that 31st Dec is part of 2018 but belongs to week 1, 2019. The calculated goal for the rest of week 1, 2019 should be checked too.

jaragunde:
Fixed by this commit, reviewed in PR #520. Mind you that the result provided is 0 and not the correct one, but we'd need more work and will be revisited in #521.
```
commit 526d37719cd4ae447c1fb81e57428de4b33f62d8
Author: Ana Rute Mendes <ana@igalia.com>
Date:   Wed Sep 29 16:39:35 2021 -0300

    Fix weekly goal calc if in last week of the year
    
    If the week in case is the last week of the year,
    this method would fail because weeksTillEndOfJourneyPeriod
    is 0, causing a division by 0 exception
    
    Fixes https://github.com/Igalia/phpreport/issues/432

```

PR #520, opened 2021-09-29
Fix weekly goal calc if in last week of the year

anarute:
If the week in case is the last week of the year, this method would fail because weeksTillEndOfJourneyPeriod is 0, causing a division by 0 exception.

### DIFF
--- a/model/facade/action/GetPersonalSummaryByLoginDateAction.php
+++ b/model/facade/action/GetPersonalSummaryByLoginDateAction.php
@@ -195,15 +195,15 @@ class GetPersonalSummaryByLoginDateAction extends Action{
         }
         $totalResults = $dao->getPersonalSummary($user->getId(), $this->date);
 
+        $totalResults['weekly_goal'] = 0;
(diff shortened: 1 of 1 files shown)


jaragunde (approved):
This fixes the error, but provides a result of 0 in the last week of the year which is not correct either. Still, I can't find the way to get the result I would expect!

These weekly goals are extremely cursed and it's hard to understand how they work, the feature should be reworked... But that's a problem for another day, let's fix this failure now.

This would fix #432, would you mind adding the issue number to the commit message? Thanks!

anarute:
> This fixes the error, but provides a result of 0 in the last week of the year which is not correct either. Still, I can't find the way to get the result I would expect!

I was doing some tests and I think we'll need to refactor the whole logic.. it doesn't seem to be generating the expected result.  Created #521 for this since fixing the division by zero error is a blocker for me at the moment.

> This would fix #432, would you mind adding the issue number to the commit message? Thanks!

Done

PR #522, opened 2021-10-04
Improve holiday summary

anarute:
Add pending holidays to the summary and add snack bar to show server response after update.

Screen Shot 2021-10-04 at 18 38 29

Screen Shot 2021-10-04 at 18 39 26

Depends on #518 and #520 (@jaragunde I didn't push to the same branch to make the review easier, since you have checked the others already)


jaragunde (changes requested):
This is looking great, thanks! Some comments:

* I noticed that the "pending" count doesn't refresh after saving your changes, like the "total booked" count does. Could you please fix that?
* The "total count" is not a good metric for us, because people may still input half-holidays manually and they count like 1 day here. The count is useful for many users, though, so I'd recommend to keep it and add context, like "hh:mm booked in N days". You may write this down as an enhancement for later :slightly_smiling_face:

jaragunde:
Another comment, before I forget. I tend to use my browser in half-screen width, that's 960 pixels width. In that case, the calendar "flexes" to only show half a year, but it tends to reset to the first half of the year in certain occasions, like right after you click save, or when the save confirmation popup disappears. See the video:

https://user-images.githubusercontent.com/2900877/136010041-0f627371-fe19-41df-a4ac-34e185e0882f.mp4

I wonder if these annoyances could be fixed, or maybe it's better to just force the calendar to show the entire year, even if it means adding a scrollbar?

anarute:
After the latest changes, the summary looks like this:
Screen Shot 2021-10-07 at 12 27 59

anarute:
> I wonder if these annoyances could be fixed, or maybe it's better to just force the calendar to show the entire year, even if it means adding a scrollbar?
I found a bug that was forcing to return to the beginning of the year every time, can you check now please?
I also improved the grid a bit for smaller screens

anarute:
> The "total booked" holidays should be split in "enjoyed" (holidays in the past) and "scheduled" (holidays in the future).

Just broke down total into scheduled and enjoyed.

> There is a potential problem with the way we transform holiday hours into days and hours, for those cases where a person has a mixture of different dedications along the week. I don't have a good solution for this, other than displaying the figure in hours directly.

For this one, I suggest us to create a task and work on it soon but not as a high priority, as we still have a few important features to implement for this and it doesn't seem to be a very common case, is it? What do you think?

jaragunde:
> For this one, I suggest us to create a task and work on it soon but not as a high priority, as we still have a few important features to implement for this and it doesn't seem to be a very common case, is it? What do you think?

Yes, let's do this later, but it's not super uncommon; we usually have a couple of people in this situation at any given time.

PR #531, opened 2021-11-03
Add team calendar tab

anarute:
Allow to pick a user in an autocomplete dropdown to visualize the holidays from any other person.

https://user-images.githubusercontent.com/333447/140186720-0d797c26-b990-47ee-a5da-949b651d8e9f.mov

(sorry for the long PR, I tried to break it down but the whole change is a bit connected)


jaragunde (approved):
Thank you!

I haven't reviewed the front-end thoroughly, unfortunately I've lost track of how it works at this point. Manual testing seems to be fine, let's merge so we can keep moving!

PR #532, opened 2021-11-03
Increase the number of months available in the calendar

anarute:
Especially by the end and beginning of the year, is useful to be
able to see close months from the past or next year so display
the holidays since July from past year until March of the
next year.

Needs https://github.com/Igalia/phpreport/pull/531 to be merged first.

### DIFF
--- a/model/facade/UsersFacade.php
+++ b/model/facade/UsersFacade.php
@@ -259,8 +259,8 @@ abstract class UsersFacade {
         return $action->execute();
     }
 
-    static function GetHolidayHoursSummary(DateTime $init, DateTime $end, UserVO $user = NULL) {
-        $action = new GetHolidayHoursSummaryAction($init, $end, $user);
+    static function GetHolidayHoursSummary(DateTime $init, DateTime $end, UserVO $user = NULL, Datetime $referenceDate = NULL) {
+        $action = new GetHolidayHoursSummaryAction($init, $end, $user, $referenceDate);
         return $action->execute();
     }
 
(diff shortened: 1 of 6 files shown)


jaragunde (commented):
There are several things that I find strange:
* Showing only some months of the previous and next year doesn't seem particularly useful.
* The calendar is not centered any more in the year in course, and it seems impossible to see the full year at all.
* We can edit holidays in the future, but there isn't a summary for that year.

I would prefer to shift time using the year as the minimum unit. When you click right, you'd see the entire next year (or its first half if the screen is too narrow), and when doing so the summary changes to the one of that year.

Maybe we could have more input on this topic...

anarute:
Thanks for your input, I've changed the behavior so the year always starts on Jan of the present year and managed to update the summary details when the user changes the calendar page. I think it's much better now, let me know what you think :)

jaragunde (changes requested):
I found a small problem in the team calendar:
1. Reduce the browser window until only half of the year is shown.
2. Click left to go back to the previous year. It will go straight to the first half of the year.
3. Click right to see the second half of the year, you will be moved back to H1.

This only happens in the team calendar, "my holidays" works properly.

anarute:
> I found a small problem in the team calendar:
> 
>     1. Reduce the browser window until only half of the year is shown.
> 
>     2. Click left to go back to the previous year. It will go straight to the first half of the year.
> 
>     3. Click right to see the second half of the year, you will be moved back to H1.
> 
> 
> This only happens in the team calendar, "my holidays" works properly.

Nice catch! I haven't noticed. This should be fixed now.

jaragunde (approved):
Looks and works great! Thanks.

Issue #513, opened 2021-09-16
Adopt PDO for database access


jaragunde:
PHP Data Objects (PDO) define a consistent interface for accessing databases in PHP: https://www.php.net/manual/en/intro.pdo.php.

It can access different database providers with the same interface, saving ourselves the hassle to code DB access different implementations with `pg_*` or `mysql_*` methods if we need to support multiple backends. 

More importantly, PDO provides a defense against SQL injection using parameter binding for DB queries. We currently have some protections using the DBAdapter code that we imported from the Propel project, but we must not forget to use them when coding queries and we probably failed to do so here and there.

Let's figure a way to adopt PDO incrementally in our code base!

jaragunde:
An interesting tutorial: https://www.sitepoint.com/re-introducing-pdo-the-right-way-to-access-databases-in-php/

jaragunde:
A lesson learned when coding the first patches: if we want to use `PDO::FETCH_CLASS`, we need to modify the VOs to have the internal properties match the column names.

I already attempted to match column names in the query with the internal properties in the VOs using AS like `SELECT usrid AS userId`, but it returns column names as lower case. The result is it creates a new `userid` property in `TemplateVO`, instead of assigning the correct value to the existing `userId` property.

jaragunde:
Reviewed in PR #534:
```
09a5efa6 [#513] Port Template DAO to use PDO. [Jacobo Aragunde Pérez]
9460159d Use PDO::FETCH_CLASS to retrieve TemplateVO objects. [Jacobo Aragunde Pérez]
f74cf305 Docs: Add PDO to system requirements. [Jacobo Aragunde Pérez]
01afc795 Port PostgreSQLTemplateDAO::delete() to PDO. [Jacobo Aragunde Pérez]
a4ea9bd5 Port PostgreSQLTemplateDAO::create() to PDO. [Jacobo Aragunde Pérez]

**Comment by jaragunde** [comment]
From #539:
```
f134ab10 [#513] Port Config DAO to use PDO. [Jacobo Aragunde Pérez]
81a521dd Port PostgreSQLConfigDAO to use PDO. [Jacobo Aragunde Pérez]
c868e06b Remove unused operation ConfigDAO::getVersionNumber(). [Jacobo Aragunde Pérez]
1b83e5c8 Move PDO object up in the DAO hierarchy to BaseDAO. [Jacobo Aragunde Pérez]
[...]

PR #534, opened 2021-11-24
Port Template DAO to use PDO

jaragunde:
First steps towards adopting PDO (#513), starting with an isolated data type that's not used a lot.


anarute (approved):
It looks good to me and I also tested locally. Thanks for this improvement!

Issue #434, opened 2018-12-05
Add email field to users table


jaragunde:
User emails are required by the task creation reminder feature implemented in #334, which sends those reminders via email. That was implemented by appending a certain "@domain.com" to the user login name but, of course, actual email accounts could be different, and we should support that scenario.

Add the email field in the DB, the UI to edit it, and the means for #334 to use it (also use it in #391 and #429 if they are ever implemented).

PR #535, opened 2021-11-25
Sync PhpReport holidays with CalDav Calendar

anarute:
Adds a button in the calendar view that allows the user to sync their PhpReport holidays with a CalDav Calendar.

To test and deploy this one it will be needed to run `composer install` again since we added an external [package](https://packagist.org/packages/thecsea/simple-caldav-client) to handle the CalDav requests.

There are also a few new config variables that need to be filled in order for it to work:
```
define('CALENDAR_URL', '');
define('CALENDAR_ID', '');
define('CALENDAR_USERNAME', '');
define('CALENDAR_PASSWORD', '');
define('COMPANY_DOMAIN', '');
```

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -194,10 +194,10 @@ define('MENU_COORDINATION', TRUE);
 define('EXTRA_HOURS_WARNING_TRIGGER', 50);
 
 /**
- * @name NO_FILL_EMAIL_DOMAIN
+ * @name COMPANY_DOMAIN
  * @global string Domain of the users pinged on the NO_FILL emails
  */
-define('NO_FILL_EMAIL_DOMAIN', "domain.com");
+define('COMPANY_DOMAIN', "domain.com");
 
 /**
  * @name NO_FILL_EMAIL_FROM
@@ -332,4 +332,3 @@ define('CALENDAR_URL', '');
 define('CALENDAR_ID', '');
 define('CALENDAR_USERNAME', '');
 define('CALENDAR_PASSWORD', '');
-define('COMPANY_DOMAIN', '');
--- a/config/config.php
+++ b/config/config.php
@@ -326,3 +326,10 @@ define('ISSUE_TRACKER_LINKS_URL',
  * To be used in combination with ISSUE_TRACKER_LINKS_URL, see its entry above.
  */
 define('ISSUE_TRACKER_LINKS_TEXT', serialize(array('Report an issue')));
+
+/* CalDav integration */
+define('CALENDAR_URL', '');
+define('CALENDAR_ID', '');
+define('CALENDAR_USERNAME', '');
+define('CALENDAR_PASSWORD', '');
+define('COMPANY_DOMAIN', '');
--- a/web/js/holidayManagement.js
+++ b/web/js/holidayManagement.js
@@ -341,6 +341,8 @@ var app = new Vue({
             }
         },
         onSaveClick: async function () {
+            this.toggleDisableButtons();
+
             const url = `services/updateHolidays.php?init=${formatDate(this.init)}&end=${formatDate(this.end)}`;
             const res = await fetch(url, {
                 method: 'POST',
@@ -377,8 +379,10 @@ var app = new Vue({
                     this.serverMessages.push({ classes: "message success", text: "Holidays were updated." });
                 }
             };
-            this.fetchSummary();
-            this.$emit('flush-message')
+            await this.fetchSummary();
+            this.syncCalendar();
+            this.toggleDisableButtons();
+            this.$emit('flush-message');
         },
         onSelectUser(userIndex) {
             if (!this.usersList[userIndex]) return;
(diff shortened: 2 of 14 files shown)


config/config.php
```diff
@@ -326,3 +326,10 @@ functionality to PhpReport
  * To be used in combination with ISSUE_TRACKER_LINKS_URL, see its entry above.
  */
 define('ISSUE_TRACKER_LINKS_TEXT', serialize(array('Report an issue')));
+
+/* CalDav integration */
+define('CALENDAR_URL', '');
+define('CALENDAR_ID', '');
+define('CALENDAR_USERNAME', '');
+define('CALENDAR_PASSWORD', '');
+define('COMPANY_DOMAIN', '');
```
jaragunde, line 335: We also have `define('NO_FILL_EMAIL_DOMAIN', "domain.com");` for a similar purpose: compose email addresses for our users.

The name of the existing parameter is quite specific, maybe we could make the other feature use `COMPANY_DOMAIN`? Or keep both, no strong opinion. What we would really need is to work on #434...

config/config.php
```diff
@@ -326,3 +326,10 @@ functionality to PhpReport
  * To be used in combination with ISSUE_TRACKER_LINKS_URL, see its entry above.
  */
 define('ISSUE_TRACKER_LINKS_TEXT', serialize(array('Report an issue')));
+
+/* CalDav integration */
+define('CALENDAR_URL', '');
+define('CALENDAR_ID', '');
```
jaragunde, line 332: I don't know what this parameter means or how to get it.

web/js/holidayManagement.js
```diff
@@ -430,6 +430,52 @@ var app = new Vue({
...
+            this.$refs.syncBtn.disabled = !this.$refs.syncBtn.disabled;
+            this.$refs.saveBtn.disabled = !this.$refs.saveBtn.disabled;
+        },
+        syncCalendar: async function () {
+            const url = `services/syncCalendar.php?init=${formatDate(this.init)}&end=${formatDate(this.end)}`;
+            const ranges = this.ranges.map(range => {
+                if (range.coveredDates) {
+                    const end = addDays(range.dates.end, 1);
+                    return {
+                        start: formatDate(range.dates.start),
```
jaragunde, line 444: I'm getting this error, and the request is never sent to the web service:
```
TypeError: date is undefined
    formatDate http://localhost/phpreport/web/js/holidayManagement.js:23
    ranges http://localhost/phpreport/web/js/holidayManagement.js:444
    syncCalendar http://localhost/phpreport/web/js/holidayManagement.js:440
    VueJS 30
    <anonymous> http://localhost/phpreport/web/js/holidayManagement.js:66

**Review by jaragunde — CHANGES_REQUESTED** [review]
Sorry, I'm having some trouble to make this work! See comments.

File: config/config.php (hunk 1 above)
Comment (anarute) [reply config/config.php]: Oh #434 looks really useful indeed, I'll try to prioritize it. I renamend `NO_FILL_EMAIL_DOMAIN` to `COMPANY_DOMAIN`

File: web/js/holidayManagement.js (hunk 3 above)
Comment (anarute) [reply web/js/holidayManagement.js:444]: Thanks for catching! This was hapening for partial leaves. I just committed a patch.

File: config/config.php (hunk 2 above)
Comment (anarute) [reply config/config.php:332]: This is an odd flow indeed. So even though we pass the calendar url to a specific calendar, when connecting to the calendar the server, it returns **all** the calendars for that user, that's why we need the url and the ID.

I don't know if that's the standard for every caldav server, but in our case (Sogo), the calendar ID is the last parameter of the calendar URL, for example:

From `https://calendar-beta.igalia.com/SOGo/dav/igalia-holidays/Calendar/50E03-619E9800-B9-7E0D1A80/` the ID is `50E03-619E9800-B9-7E0D1A80`.

**Review by jaragunde — CHANGES_REQUESTED** [review]
Thanks for your work! Just one more thing...

I think the syncCalendar operation should fail gracefully when caldav sync is not configured. It currently fails with an error 500, I think it's because of:
```
[08-Dec-2021 08:21:56 UTC] PHP Fatal error:  Invalid URL: '' in /var/www/html/phpreport/vendor/thecsea/simple-caldav-client/src/CalDAVClient.php on line 120
```

anarute:
> Thanks for your work! Just one more thing...
> 
> I think the syncCalendar operation should fail gracefully when caldav sync is not configured. It currently fails with an error 500, I think it's because of:
> 
> ```
> [08-Dec-2021 08:21:56 UTC] PHP Fatal error:  Invalid URL: '' in /var/www/html/phpreport/vendor/thecsea/simple-caldav-client/src/CalDAVClient.php on line 120
> ```

I think it should fail with 500 because we made a request to something missconfigured, so if the calendar variables are blank it will return this message: "CalDAV calendar not configured correctly"

jaragunde (approved):
LET'S GOOO! Thanks!

PR #536, opened 2021-12-07
Rename `enjoyed` to `used` when referring to past holidays

anarute:
Not every taken leave was really "enjoyed"/for pleasure

### DIFF
--- a/docs/user/tasks.rst
+++ b/docs/user/tasks.rst
@@ -169,7 +169,7 @@ The *User work summary* panel on the left contains the following information:
 
 * **Extra (year):** Number of extra hours corresponding to the current year.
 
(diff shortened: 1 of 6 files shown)


joanmarie:
"Used" is a good choice. Thanks! :slightly_smiling_face:

jaragunde (changes requested):
Thanks! But please, check the tasks page; there's another "enjoyed" there (which is now blank because it cannot find the corresponding tag in the results obtained from the server).

anarute:
> Thanks! But please, check the tasks page; there's another "enjoyed" there (which is now blank because it cannot find the corresponding tag in the results obtained from the server).

I have changed [there](web/js/tasks.js) too, can you check if it can be cache on your browser?

jaragunde (approved):
My bad, sorry!

PR #539, opened 2021-12-15
Port Settings DAO to use PDO

jaragunde:
This does some refactoring to reuse the PDO connection object and ports another DAO, ConfigDAO, to use this API.

Part of #513.


anarute (approved):
It looks good to me, I also tested locally.

PR #540, opened 2021-12-15
Remove unused entities CustomEvent and ProjectSchedule

jaragunde:
This is code related with a feature that existed in PhpReport 1.x and that was never ported over to 2.x. After 12 years, we should probably accept that it won't ever be implemented and move on, removing the unused code.

Data tables are still in the database, like tables for other entities that were removed before. We will have to do a cleanup at some point, but there are other things that might be removed mid-term.


anarute (approved):
It looks good to me, always good to do some clean up :)

Issue #542, opened 2021-12-17
Filter only active users in holiday team calendar


jaragunde:
The user selector for the "Team calendar" inside the holidays page lists all the users in the system.

It can get hectic in long-time deployments, so I propose to include only active users. We are doing something similar in other places, e.g. the accumulated hours report lists only active users.

anarute:
@jaragunde I'm working on this and I've noticed that for the accumulated hours report the way to detect active users is by getting the users from the group `ALL_USERS_GROUP`, right?
What do you think of adding a boolean field to flag active fields directly in the users table?
I also miss having the user full name and email in the table, or maybe a profile table with personal information.

Anyway I'm fixing this issue by using the same approach as the report but if you agree on adding to the database I can create an issue to work on it soon.

jaragunde:
So we have a little mess in the users/groups relation. In a DB-backed phpreport, groups are a table in the database, but in an LDAP-backed phpreport, groups don't exist in the database but in LDAP. That's why we have variables in config.php to define the list of groups that matter to the system (`USER_GROUPS`), or which is the group that defines active users (`ALL_USERS_GROUP`).

If we added a flag then we'd have another piece of data to sync between LDAP and DB. Maybe it's better to create the operation "get me the list of active users" in a DAO so the upper layers wouldn't have to know about how it's done or check the application config. Then, we could eventually reimplement it like a query "get me the list of users that belong to some DB group" or something equivalent in LDAP terms.

Our original intention was that groups could be very customizable, that every deployment could configure their custom set of groups and which parts of the application they have permission into (hence `config/permissions.php`). I regret that decision and now I wish we had something simpler, just hard-coded roles instead of configurable groups, which would define levels of access, and a configurable mapping between users and roles (like users->groups now), or between LDAP groups and roles.

> I also miss having the user full name and email in the table, or maybe a profile table with personal information.

I miss that too. The full name could be used in reports and the email in things like calendar sync or email reminders. There's already a ticket for the latter (#434).

jaragunde:
Reviewed in #544:
```
commit 9be6345ba440559012214c639f051558bb4f1033
Merge: 215a46a5 6f8852b4
Author: Jacobo Aragunde Pérez <jaragunde@igalia.com>
Date:   Tue Jan 4 05:57:22 2022 +0100

    [#542] Filter active users in the Team calendar
    
    Merge pull request #544 from Igalia/filter-active-users

[...]

PR #544, opened 2021-12-20
Filter active users in the Team calendar

anarute:
List only users that belong to `ALL_USERS_GROUP`. This means that
they are active in the company and therefore this will hide inactive users.

Fixes https://github.com/Igalia/phpreport/issues/542


jaragunde (approved):
Looks good to me! Do you want to land already, or do something else in the line of what was discussed in the ticket? Feel free to merge yourself while I'm away.

anarute:
> Looks good to me! Do you want to land already, or do something else in the line of what was discussed in the ticket? Feel free to merge yourself while I'm away.

I've created an action that returns only the active users. If we add the flag in the future we can just change it :)
Let me know what you think

jaragunde (approved):
Looks great! Thanks.

PR #546, opened 2022-01-03
Fix calDav event description

anarute:
Change how to invite the user to the event so the event is created in the holidays calendar but also displayed in the user's personal calendar.


jaragunde (approved):
LGTM, and works as intended. Thanks!

Issue #543, opened 2021-12-17
Part-time holidays listed as "1 day"


jaragunde:
In the list of "days booked per week", in the holidays page, every day with any holidays is counted as one day, even if it's only a few hours reported as holidays and not an entire day. It can be a bit confusing.

PR #548, opened 2022-01-06
Consider partial leaves when grouping by week

anarute:
Instead of counting as a full day every day that has a leave,
proportionate when there are partial leaves in the week.

This was causing some confusion because the sum was always
considering each day that had a leave as 1 and creating
unprecise numbers.

Note that one day is the user journey at the time the page was
rendered.

Fixes https://github.com/Igalia/phpreport/issues/543


anarute:
Sorry I forgot we had automated tests for this, fixing them.

anarute:
Tests fixed and passing :)

jaragunde (approved):
Looks great, thanks!

PR #549, opened 2022-01-07
Add gh action to run tests

anarute:
Add action to be triggered on every pull request and run the automated tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "igalia/phpreport",
     "description": "phpreport",
     "type": "project",
-    "license": "GNU GPLv3",
+    "license": "GPL-3.0-or-later",
     "autoload": {
         "psr-4": {
             "Phpreport\\Model\\": "model/",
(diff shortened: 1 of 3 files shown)


anarute:
@jaragunde now the automated tests will run on every PR 🎉

composer.json
```diff
@@ -2,7 +2,7 @@
     "name": "igalia/phpreport",
     "description": "phpreport",
     "type": "project",
-    "license": "GNU GPLv3",
+    "license": "LGPL-3.0-or-later",
```
jaragunde, line 5: There are a few imported lines of LGPL code, but the whole application is distributed under GPL terms. Is there a problem with "GPL-3.0-or-later"?

jaragunde (commented):
This is a great step towards code reliability, thanks! I just have a nit regarding the license.

composer.json
anarute: No problem, I've just changed it to GPL-3.0-or-later

PR #550, opened 2022-01-07
Create Holiday Summary report with CSV download support

anarute:
- Add view to display the holiday summary for all active users
- Add service to download the report

The data in this report is currently managed manually with the aggregation of several other reports, so this report is intended to automate this job.

Screen Shot 2022-01-07 at 15 31 32


anarute:
Depends on #548

web/services/downloadHolidaySummary.php
```diff
@@ -0,0 +1,67 @@
...
+
+$fp = fopen('php://output', 'wb');
+
+$today = date_create();
+fputcsv($fp, array('Report generated at', $today->format('Y-m-d H:i:s e')), ';');
+fputcsv($fp, array(''), ';');
+
+$weeksLine = array_merge(array('User', 'Hours/day', 'Holidays(days)', 'Holidays(hours)', 'Pending (hours)', 'Planned (hours)', '% planned'), array_keys($usersAndWeeks["weeks"]));
+
+fputcsv($fp, $weeksLine, ';');
```
jaragunde, line 45: Just curious, why the semicolon as separator? I think we've been using commas in other places. fputcsv will enclose strings in `""` if they contain the separator character.

config/permissions.php
```diff
@@ -49,6 +49,7 @@
         '/services/syncCalendar.php',
         '/holidayManagement.php',
         '/holidaySummary.php',
+        '/services/downloadHolidaySummary.php',
```
jaragunde, line 52: I originally wanted to have services admit a `format` parameter, so we would use `?format=xml` or `?format=csv` to get the same set of data in different formats. We even implemented it in some services. Not a strong opinion, as the codebases is a bit of a mess right now, specially the services layer.

jaragunde (commented):
I have some comments. Besides, are you considering to add a way to select the year? That will probably be necessary as we approach the end of 2022.

web/services/HolidayService.php
```diff
@@ -150,6 +150,23 @@ static function formatHours(float $time, float $journey, int $limit): string
...
+    static function getWeeksFromYear($year = NULL): array
+    {
+        $weeks = [];
+        $year = $year ?? date('Y');
+        $first_week = date('W', mktime(0, 0, 0, 1, 1, $year));
+        $last_week = date('W', mktime(0, 0, 0, 12, 31, $year));
+        if ($first_week == "52" || $first_week == "53") {
+            $weeks[($year - 1) . "W" . $first_week] = 0;
+            $first_week = "1";
+        }
```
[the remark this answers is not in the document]
anarute: Nice catch. Just added a logic for that

anarute:
> I have some comments. Besides, are you considering to add a way to select the year? That will probably be necessary as we approach the end of 2022.

Tbh, I haven't thought about that. I'm hoping that this year we can setup a business inteligence tool (like grafana or metabase) and start doing reports there instead of creating more reports inside PhpReport. I've been experimenting with [Metabase](https://www.metabase.com/) and enjoying it a lot!

web/services/downloadHolidaySummary.php
anarute: No particular reason, probably copied from somewhere else :)

I've just switched to comma.

config/permissions.php
anarute: Right, it makes sense to merge them, I just did that.

web/services/HolidayService.php
```diff
@@ -330,4 +330,64 @@ public function syncCalDAVCalendar(array $datesRanges = [])
...
+                'availableDays' => $this::formatHours($summary['availableHours'][$users[$i]->getLogin()], $validJourney, 1),
+                'pendingHours' => round($summary['pendingHours'][$users[$i]->getLogin()], 2),
+                'usedHours' => round($summary['usedHours'][$users[$i]->getLogin()], 2),
+                'percentage' =>  $summary['availableHours'][$users[$i]->getLogin()] ? round(($summary['usedHours'][$users[$i]->getLogin()] / $summary['availableHours'][$users[$i]->getLogin()]) * 100, 2) : 0,
+                'hoursDay' => $validJourney,
+                'holidays' => $leaves,
+            ];
+        }
+        asort($holidays);
+        return [
```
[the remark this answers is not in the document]
anarute: True, I've moved it to an action: [model/facade/action/GetHolidaySummaryReportAction.php](https://github.com/Igalia/phpreport/pull/550/files#diff-133cddd0577090bfd58ebe45aca8040caffc3e1b1e5168c569cfe44f9c3f25a1)

web/services/getHolidaySummary.php
```diff
@@ -0,0 +1,74 @@
...
+require_once(PHPREPORT_ROOT . "/vendor/autoload.php");
+require_once(PHPREPORT_ROOT . '/util/LoginManager.php');
+
+$csvExport = array_key_exists("format", $_GET) && $_GET["format"] == "csv";
+
+$loginManager = new \LoginManager();
+
+$holidayService = new HolidayService($loginManager);
+
+$usersAndWeeks = $holidayService->retrieveHolidaysSummary();
```
jaragunde, line 34: I passed "2024" as a parameter here, to test the results of a different year, and I'm still getting 2022 results. Can you check why?

jaragunde (changes requested):

web/services/getHolidaySummary.php
anarute: The year wasn't being passed to the method that generates the weeks per year. Can you try now, please?

I also added support to the service endpoint: `/web/services/getHolidaySummary.php?year=2024`

jaragunde: Awesome, thanks!

jaragunde (approved):
Good job, thank you!

PR #551, opened 2022-01-12
Port Customer DAO to use PDO

jaragunde:
This does some refactoring to reuse the `runSelectQuery()` function and ports another DAO, `CustomerDAO`, to use the PDO API. Part of #513.

Includes a patch for #431. Also removes unused code.


web/services/getUserCustomersService.php
```diff
@@ -31,23 +31,17 @@
...
     if (strtolower($active) == "true")
-    $active = True;
+        $active = True;
     else
-    $active = False;
-
-    $login = $_GET['login'];
-
-    $sid = $_GET['sid'];
+        $active = False;
```
anarute, line 40: This is nitpick, but I like removing the Else when possible for a cleaner code, like:

```
    $active = False;
    if (strtolower($active) == "true")
        $active = True;
```

model/dao/CustomerDAO/PostgreSQLCustomerDAO.php
```diff
@@ -127,12 +127,18 @@ public function getBySectorId($sectorId, $orderField = 'id') {
...
     public function getByProjectUserLogin($userLogin, $active = False, $orderField = 'id') {
-        $sql = "SELECT * FROM customer WHERE id IN (SELECT customerid FROM requests WHERE projectid IN (SELECT id FROM project WHERE";
         if ($active)
-            $sql = $sql . " activation = 'True' AND";
-        $sql = $sql . " id IN (SELECT projectid FROM project_usr WHERE usrid = ( SELECT id FROM usr WHERE login = " . DBPostgres::checkStringNull($userLogin) . " )))) ORDER BY " . $orderField  . " ASC";
-        $result = $this->execute($sql);
-        return $result;
+            $activeCondition = "activation = 'True'";
+        else
+            $activeCondition = "TRUE";
```
anarute, line 102: Nit: removing the else block

```
        $activeCondition = "TRUE";
        if ($active)
            $activeCondition = "activation = 'True'";
```

anarute (approved):
Clean up looks good, I haven't tested the code manually though and added some nit comments but it's a personal preference, if you don't agree with them you can ignore

PR #552, opened 2022-01-14
Fix null access to $oldTask.

jaragunde:
In case there was no task for the specified ID, the code would break. Now we check if it exists before actually using the returning object.


jaragunde:
Thanks for taking a look!

PR #554, opened 2022-01-17
Ignore holiday views and vendor folder on Makefile

anarute:
Avoid trying to run on composer libs and replacing VueJs files


jaragunde (approved):
I gave it a try, looks good to me. Thanks!

PR #555, opened 2022-01-25
Fix more PHP warnings

jaragunde:
I went through the server logs and fixed a pile of these.

### DIFF
--- a/web/services/getUserProjectCustomerReportJsonService.php
+++ b/web/services/getUserProjectCustomerReportJsonService.php
@@ -35,19 +35,17 @@
 
     $userId = $_GET['uid'];
 
-    $init = $_GET['init'];
+    $init = $_GET['init'] ?? "";
 
-    $end = $_GET['end'];
+    $end = $_GET['end'] ?? "";
 
-    $dateFormat = $_GET['dateFormat'];
+    $dateFormat = $_GET['dateFormat'] ?? "Y-m-d";
 
-    $login = $_GET['login'];
-
-    $sid = $_GET['sid'];
+    $sid = $_GET['sid'] ?? NULL;
 
     do {
 
-        $response[uid] = $userId;
+        $response['uid'] = $userId;
 
         /* We check authentication and authorization */
         require_once(PHPREPORT_ROOT . '/util/LoginManager.php');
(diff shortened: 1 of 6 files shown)


web/services/getUserProjectCustomerReportJsonService.php
```diff
@@ -35,67 +35,62 @@
 
     $userId = $_GET['uid'];
 
-    $init = $_GET['init'];
+    $init = $_GET['init'] ?? "";
```
anarute, line 38: Any reason why not using `NULL` instead of `""`? I think it makes more sense

web/services/getUserProjectCustomerReportJsonService.php
```diff
@@ -35,67 +35,62 @@
...
 
-    $end = $_GET['end'];
+    $end = $_GET['end'] ?? "";
```
anarute, line 40: Same here

web/services/getUserProjectCustomerReportJsonService.php
```diff
@@ -142,53 +139,53 @@
...
 
             $records[] = $record;
         }
 
         foreach($records as $record)
         {
-            $record[percentage] = round(100 * $totalHours[$record[project]]/$totalHours[total], 2, PHP_ROUND_HALF_DOWN);
-            $response[records][] = $record;
+            $record['percentage'] = round(100 * $totalHours[$record['project']]/$totalHours['total'], 2, PHP_ROUND_HALF_DOWN);
+            $response['records'][] = $record;
```
anarute, line 157: I wonder how this was working before 😅

anarute (approved):

web/services/getUserProjectCustomerReportJsonService.php
jaragunde: Hehe, PHP is extremely permissive. But it will stop working at some point, according to the error message:
```
[php7:warn] [pid 23627] PHP Warning:  Use of undefined constant sortable - assumed 'sortable' (this will throw an Error in a future version of PHP)

web/services/getUserProjectCustomerReportJsonService.php
jaragunde: I keep it like that because, later in the code, it checks for "" and not null (`if ($init!="")...`. I only wanted to get rid of the warning but not change a lot of code.

This kind of check is repeated over and over in the web services code, if I recall correctly, the original intention was to check for an empty parameter like in `getUserProjectCustomerReportJsonService.php?init=&...`.

anarute: Oh I see, that's fine then, thanks

jaragunde: Indeed, I checked and if you pass an empty GET parameter its value will be set as an empty string. I think it's a reasonable default.

jaragunde:
Thanks for taking a look!

PR #556, opened 2022-01-26
Remove project coordination features.

jaragunde:
They haven't been actively developed for years and we don't know of any deployment using them.

We are not altering the DB structure or data yet, any data will be preserved.

I would have liked to do this in smaller steps, but once I started deleting code, things were breaking because they are quite inter-dependent, and I ended up doing one big commit.


model/dao/TaskDAO/PostgreSQLTaskDAO.php
```diff
@@ -968,7 +921,7 @@ public function update(TaskVO $taskVO) {
     public function create(TaskVO $taskVO) {
         $affectedRows = 0;
 
-        $sql = "INSERT INTO task (_date, init, _end, story, telework, onsite, text, ttype, phase, usrid, projectid, task_storyid) VALUES(" .
+        $sql = "INSERT INTO task (_date, init, _end, story, telework, onsite, text, ttype, phase, usrid, projectid) VALUES(" .
```
anarute, line 924: What's the difference between story and task story id?

jaragunde: In the "Coordination" menu, you used to be able to create iterations and stories inside of them, which contained tasks (named task story). Tasks could be assigned to a task inside a story through the task story id field. The tasks UI was able to list the open stories for a certain project to select among them.

Story is a text identifier that was originally used to relate tasks with stories in an external coordination tool. It's still useful as a free-form identifier to classify tasks (there's a report for that), that's why I'm keeping this one and not the other.

anarute: Got it, thank you

anarute (approved):

jaragunde:
I'm a bit nervous about this, but here we go! :crossed_fingers:

anarute:
🚀 🚀 🚀 🚀 

I'll do some tests locally and we can deploy to beta next Monday

PR #557, opened 2022-01-31
Rename Holiday to Vacation

anarute:
Following the new agreement, we should start using Vacation.

@jaragunde I haven't changed the project name as this would need a change in config and in the db, right? Not sure if it's worth do do it.

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -1098,7 +1098,7 @@ Ext.onReady(function(){
             name: 'holidays_title',
             labelSeparator: '',
             labelStyle: userWorkSummaryTitleStyle,
-            fieldLabel:'Holidays',
+            fieldLabel:'Vacation days',
         },
         {
             id:'available_holidays',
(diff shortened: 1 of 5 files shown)


jaragunde:
I'd do the change because the project name is quite visible, although now we have the holidays feature we might not use "holiday tasks" that much.

It should be enough to change the project name plus this bit of `config.php`:
```php
define('VACATIONS_PROJECT', 'Holidays');
```
Honestly, it should have been an ID...

jaragunde (approved):
Looks good to me, but I'd welcome another eye from the "editorial" point of view. Thanks in advance!

web/js/tasks.js
```diff
@@ -1193,7 +1193,7 @@ Ext.onReady(function(){
             // Default button for full-holiday task
             new Ext.Button({
                 id: 'fullHolidayTaskButton',
-                text: 'Full-Holiday task',
+                text: 'Full Vacation Day task',
```
mrobinson, line 1196:
```suggestion
                text: 'Full vacation day task',
```
Suggesting small caps here for consistency with the other buttons.

PR #558, opened 2022-02-01
Add employees filter on Vacation Management

anarute:
- Add generic support to filter employees in `GetAllActiveUsers` façade
- Filter employees on Vacation management view

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -167,6 +167,12 @@ define ('YEARLY_HOLIDAY_HOURS', 184);
 define('ALL_USERS_GROUP', 'staff');
 
 /**
+ * @name EMPLOYEES_GROUP
+ * @global string users group used for retrieving all active employees
+ */
+define('EMPLOYEES_GROUP', 'staff');
+
+/**
  * @name USER_GROUPS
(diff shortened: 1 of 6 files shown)


config/config.php
```diff
@@ -166,6 +166,12 @@ functionality to PhpReport
  */
 define('ALL_USERS_GROUP', 'staff');
 
+/**
+ * @name EMPLOYEES_GROUP
+ * @global string users group used for retrieving all active employees
+ */
+define('EMPLOYEES_GROUP', 'employee');
```
jaragunde, line 173: I would prefer that the default value was 'staff', because 'employee' does not match any of the groups in a default install. It's a bit annoying for development purposes.

jaragunde: I mean, what's annoying is that the "Team calendar" in vacation management would not work because no users would be listed.

jaragunde (approved):
Looks good to me with a nit, see comments above.

config/config.php
anarute: ok no problema, I just changed to 'staff'

jaragunde:
Great, thanks!

PR #559, opened 2022-02-01
Add area column to vacation report

anarute:
- Create GetAreaById action
- Display only employees on Vacation Summary report
- Add User Area column on Vacation Summary report


util/DateOperations.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PhpReport.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace Phpreport\Util;
+
+class DateOperations
+{
+    public static function dateBelongsToPeriod(\DateTime $date, \DateTime $init, \DateTime $end): bool
```
jaragunde, line 26: Some time ago we added `JourneyHistoryVO:: dateBelongsToJourney`. We [mentioned](https://github.com/Igalia/phpreport/pull/515#pullrequestreview-762029898) that it would be useful to move it to the parent class `BaseHistoryVO`. Doing that, we wouldn't need this new class...

anarute: That's true, I completely forgot about that. Let me move to `BaseHistoryVO` instead

jaragunde (commented):
Looks good to me and seems to work properly. I have one comment, though.

Also, I foresee that users will want to check the data in this report sorted by areas. I propose two enhancements (not mutually excluding):

1. Provide the data grid sorted by area.
2. Use some kind of widget that allows a sortable grid. One of the good things that Ext JS had was that it provided exactly this behavior.

Maybe we could implement 1 already?

util/DateOperations.php
anarute: Done

anarute:
> Looks good to me and seems to work properly. I have one comment, though.
> 
> Also, I foresee that users will want to check the data in this report sorted by areas. I propose two enhancements (not mutually excluding):
> 
> 1. Provide the data grid sorted by area.
> 2. Use some kind of widget that allows a sortable grid. One of the good things that Ext JS had was that it provided exactly this behavior.
> 
> Maybe we could implement 1 already?

I'm currently implementing a filter per project, maybe I can add it for area too. About sorting the grid, I'll check how much time I would spend on it, but I think I will leave it for later since we want to start testing Metabase, it has all of those features implemented and I think it's better to spend time porting the reports there.

jaragunde (approved):
Looks good to me, thanks!

Issue #445, opened 2019-02-14
Multi-user project dedication report


jaragunde:
This report would show the time devoted by some users to every project during a period of time. Users would be listed in rows, and projects in columns:
```
|           | PhpReport development | Server maintenance |
| jaragunde |           100         |        0           |
| pabelenda |            35         |        300         |
```

The report would provide the following filters:

* Date filters, as usual in most reports.
* A multi-selection field for users, where you can manually write who will be listed.
* A project selector. It would populate the multi-selection field with project members.
* An area selector. It would populate the multi-selection field with are members at that moment.

The date filters are currently implemented with the `DateIntervalForm` widget, which might be reused. The user filters defined should ideally be implemented in a way they can be reused, for example, to implement #430.

Finally, this new report would replace both "users evaluation" and "projects summary" reports. Check if part of their backend operations can be reused to implement this report.

Originally from #331.

PR #561, opened 2022-02-08
Add project filter to Vacation summary report

anarute:
Add an autocomplete input that allows the user to filter the vacation report based on the projects people are assigned to.

@jaragunde I mentioned on the other PR that I was planning to add Area filter, but I've already spent way too much time on the whole vacation management feature, so I think I'll leave for later, as the project filter was the last request from the PM team for this report.


web/services/getHolidaySummary.php
```diff
@@ -27,6 +27,7 @@
 
 $csvExport = array_key_exists("format", $_GET) && $_GET["format"] == "csv";
 $year = $_GET["year"] ?? NULL;
+$users = $_GET["users"] ?? NULL;
```
jaragunde, line 30: I love how this could be combined with different kinds of front-end widgets to select a list of users from an area, a project, etc. I've been foreseeing something like this for ages, like in https://github.com/Igalia/phpreport/issues/445.

jaragunde (approved):
Thanks, it looks great!

PR #565, opened 2022-02-18
Remove WorksDAO

jaragunde:
This was a leftover of a feature in PhpReport 1.x that was never brought over to 2.x, the same as the Project Schedule DAO we already removed.


anarute (approved):

jaragunde:
Thanks!

PR #567, opened 2022-02-24
Vacation Summary: Add Used column

anarute:
As per requests from users:
- add used column based on the date the report is generated
- reorder columns
- rename Planning to Scheduled to match the User Summary labels


jaragunde (approved):
Looks good to me, thanks!

PR #568, opened 2022-03-03
Add filter by users to user management screen.

jaragunde:
Added a combo box that can be used to filter the list of users, it is very useful when the number of users is very high.

The interaction is not perfect, e.g. I don't like how the combo opens up automatically on focus, but the functionality is great IMO.


anarute (approved):
Looks good to me :)

PR #569, opened 2022-03-07
Improve usability of user management screen

jaragunde:
Some little tweaks to make this screen more pleasant to use:
* Move the recently added filter into the toolbar, which is less ugly.
* When users select a name from the filter, the corresponding row will be selected.
* The grid at the bottom will be expanded when a user is selected.
* The order of the panels was changed to better reflect their use.


anarute (approved):
Looks good to me

PR #570, opened 2022-03-07
Project management: add project name filter.

jaragunde:
Add a combo box to filter the list of projects.

~~TODO: the box should allow to search by customer name too, like we do in the tasks screen.~~

### DIFF
--- a/web/js/projectManagement.js
+++ b/web/js/projectManagement.js
@@ -315,7 +315,7 @@ Ext.onReady(function(){
                 name: 'projectFilter',
                 xtype: 'combo',
                 autoSelect: true,
-                typeAhead: true,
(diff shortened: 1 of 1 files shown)


anarute (approved):

jaragunde:
I added one new commit, please take another look. Thanks!

anarute:
> I added one new commit, please take another look. Thanks!

Looks good to me! I saw the comment about the duplicated code but not sure if it's worth to spend much time tweaking the Ext JS code since we plan to replace it at some point

jaragunde:
Sure, let's leave the duplicate code "noted" but move on. There must be some balance between cleanness and practicality!

PR #571, opened 2022-03-14
Speed up tasks page

### LINKED ISSUE #506: 3 second before interface usability when viewing a day
When viewing and editing a particular day there is a significant delay when showing pre-existing tasks or having the ability to add new tasks. From the looks of it, the UI is waiting for something before firing off the necessary XMLHttpRequests. It's likely that these should be triggered ASAP during page loading.

Screenshot from 2021-08-11 17-55-19

jaragunde:
Speed up the load of the list of tasks, by fetching them before the onReady event.

Fixes #506.

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -723,6 +723,38 @@ var TaskPanel = Ext.extend(Ext.Panel, {
     }
 });
 
+/* Proxy to load the personal summary */
+var summaryProxy = new Ext.data.HttpProxy({
+method: 'GET',
+    api: {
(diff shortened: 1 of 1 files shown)


web/js/tasks.js
```diff
@@ -221,7 +221,7 @@ function simplifyTitle(p){
  * Checks if there are unsaved changes in this page.
  */
 function isUnsaved() {
-    modifiedRecords = myStore.getModifiedRecords();
```
anarute, line 224: Much better naming 😄

anarute (approved):
looks good to me!

PR #572, opened 2022-03-16
Improve partial days text in vacation managementt

mrobinson:
Improve the display of partial vacation days. "Leave" is a broader
category of PTO than "vacation," so using "vacation" here adds
some clarity.


jaragunde (approved):
Looks good to me, thanks!

PR #573, opened 2022-03-18
Properly escape user-entered string data

### LINKED ISSUE #564: Unescaped area name can break users page
If you create an area in *Data Management -> Areas* with a name that contains the `'` character, like "I don't know", it will later break the user edition page (*Data Management -> Users*. It will be empty and show this error in the browser console:

> Uncaught SyntaxError: missing ] after element list

The reason is that, at some point, we generate a piece of JS code with a list of areas like this:
```
data : [
    [2, 'Web'],[3, 'Systems'],[4, 'Management'],[5, 'I don't know'],]
```

Where we use `'` to enclose strings, and we don't do any escaping of the string so we are generating incorrect JavaScript.

jaragunde:
Properly escape project, area, sector and customer names and other string data entered by users, in the XML and JS we generate.

Fixes #564.

### DIFF
--- a/web/projectDetailsReport.php
+++ b/web/projectDetailsReport.php
@@ -57,9 +57,9 @@ if(!LoginManager::hasExtraPermissions()) {
 echo "<!-- Global variables extracted from the PHP side -->\n";
 echo "<script>\n";
 echo "var projectData = {\n";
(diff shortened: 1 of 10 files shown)


anarute (approved):
LGTM

PR #574, opened 2022-03-18
Enhance how we fetch the vacations project id

jaragunde:
I started fixing a PHP warning and ended doing a bit of refactoring and optimization.

I extracted the code that calculates the vacations project id to its own function and reused it from `getVacations` and `getVacationsDates`. This also has the advantage of reducing the usage of the `VACATIONS_PROJECT` config parameter in case we want to get rid of it. We also cache the value for subsequent calls, it's useful for `getVacations` which we call several times in the same request.


anarute (approved):
LGTM

PR #576, opened 2022-04-08
Remove unused code in Task DAO

jaragunde:
In preparation for the PDO migration, remove as much unused code as possible so we don't have to migrate it.


anarute (approved):
LGTM

PR #579, opened 2022-04-26
Copy result of legacy CSV exporter to clipboard.

jaragunde:
Apparently, the data URLs generated by the legacy CSV exporter are not working in any browser.

Instead of trying to open a data URL, a quick workaround is to copy the contents of the CSV report to the clipboard. Users would be able to paste it into a text file, or even directly into a spreadsheet.

Copying to the clipboard appears to be more reliable than what we have now.

### DIFF
--- a/web/js/include/ExportableGridPanel.js
+++ b/web/js/include/ExportableGridPanel.js
@@ -59,18 +59,15 @@ Ext.ux.ExportableGridPanel = Ext.extend(Ext.grid.GridPanel, {
 
     initComponent: function () {
         Ext.apply(this, {
-            bbar: [
-                {
-                    xtype: 'button',
-                    text: 'Download as CSV',
-                    handler: function () {
-                        //FIXME there must be a better way to get the grid
-                        var gridComponent = this.ownerCt.ownerCt;
-                        window.open("data:text/plain," + encodeURI(
-                                fromStoreToCSV(gridComponent.getStore(), gridComponent.getColumnModel())));
-                    }
-                },
-            ],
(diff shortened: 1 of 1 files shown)


anarute:
Before approving, I tried to fix the download to keep the expected result instead of copying to the clipboard and this seems to do the trick:

```diff
diff --git a/web/js/include/ExportableGridPanel.js b/web/js/include/ExportableGridPanel.js
index 74be03f..e121b81 100644
--- a/web/js/include/ExportableGridPanel.js
+++ b/web/js/include/ExportableGridPanel.js
@@ -66,8 +66,12 @@ Ext.ux.ExportableGridPanel = Ext.extend(Ext.grid.GridPanel, {
                     handler: function () {
                         //FIXME there must be a better way to get the grid
                         var gridComponent = this.ownerCt.ownerCt;
-                        window.open("data:text/plain," + encodeURI(
-                                fromStoreToCSV(gridComponent.getStore(), gridComponent.getColumnModel())));
+                        const linkSource = `data:text/plain;${encodeURI(
+                            fromStoreToCSV(gridComponent.getStore(), gridComponent.getColumnModel()))}`;
+                        const downloadLink = document.createElement("a");
+                        downloadLink.href = linkSource;
+                        downloadLink.download = `${this.ownerCt.ownerCt.title.replace(/ /g, '')}.csv`;
+                        downloadLink.click();
                     }
                 },
             ],
```

What do you think?

jaragunde:
That's much better! Closing in favor of Ana's proposal.

PR #581, opened 2022-04-28
Deprecate VACATIONS_PROJECT config parameter

jaragunde:
Create a new parameter, VACATIONS_PROJECT_ID, and add a migration path where the old parameter will still work if the new is not set.


anarute (approved):
LGTM! To deploy this one we just need to change the vacations project ID, right?

jaragunde:
We could land it with VACATIONS_PROJECT_ID commented out or removed, then set it to the correct project id (it can be found in the project details page, for example).

PR #582, opened 2022-04-28
Remove timeout for new, empty task creation.

jaragunde:
The new, empty task that appears when no task are saved for the day feels too slow. The reason is: if the function is called before tasks have been loaded, it sets a 1-second timeout before retrying, which is too slow.

To fix the issue, we refactor the code a bit so the function is called as soon as possible, exactly when task load is done.

CC @mrobinson, this started feeling slow after landing #571 :grin:

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -287,6 +287,40 @@ function addTask(newRecord) {
     return taskPanel;
 }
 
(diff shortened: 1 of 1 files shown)


anarute (approved):
LGTM

PR #583, opened 2022-05-02
Review panel height values in user management page.

jaragunde:
There were problems with certain versions of Firefox.


anarute (approved):
Thank you!

PR #584, opened 2022-05-02
[#439] Append a hash to minified file names.

### LINKED ISSUE #439: Force browser to clear cache after upgrade
When we upgrade (or hotfix) the server, users often have trouble with cached files. There are methods to force that from the server, like appending a GET parameter to .js and .css requests with the version number, like: `ext.min.js?v=2.18`.

The method is described here: https://stackoverflow.com/a/1922924

jaragunde:
This will enforce browsers to retrieve again the file when the hash changes, making sure users don't use old versions of the JS code.

The minify script is changed in several ways, in addition to calculating and appending the hash to the file name:

* To revert previous minification, we now use git reset --hard. It's   dangerous but prevents the regexp logic from getting too wild.

* We now loop through the php files for every js file, making the script much less efficient. Again, this is done to simplify regexp  logic.

* Explicitly delete source map files from previous minification, to prevent maps with different hashes to pile up.

Fixes #439.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,21 +39,25 @@ help:
 	rm footer
 
 minify:
+	# WARNING: this will remove any unstaged changes! Do not run on a development directory
 	for i in `find -name "*.min.js" -not -path "*web/vuejs/*"`; do rm $$i; done
-	for i in `find -name "*.js" -not -path "*web/vuejs/*"` ; do \
+	for i in `find -name "*.min.js.map" -not -path "*web/vuejs/*"`; do rm $$i; done
+	#revert any previous minification changes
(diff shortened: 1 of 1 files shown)


anarute (approved):
Thanks for looking into this, I never found the time to do it.

jaragunde:
Same as I, but I found [an article](https://javascript.plainenglish.io/what-is-cache-busting-55366b3ac022) about a solution that looked straightforward to implement and I gave it a try. It wasn't so easy and I got in trouble with regexps, but I think it works in the end. We need to be careful when deploying it.

PR #588, opened 2022-06-08
Round vacation summary per week

anarute:
Make sure we don't end up with lots of decimal places breaking the summary report table.

### DIFF
--- a/web/services/HolidayService.php
+++ b/web/services/HolidayService.php
@@ -188,6 +188,7 @@ class HolidayService
                 $weeks[$current_week] = $leavesDetails[$dates[$i]]['amount'] ?? 1;
                 $previous_week = $current_week;
             }
+            $weeks[$current_week] = round($weeks[$current_week], 2);
         }
         return $weeks;
     }


web/services/HolidayService.php
```diff
@@ -188,6 +188,7 @@ static function groupByWeeks(array $leavesDetails, $weeks = []): array
                 $weeks[$current_week] = $leavesDetails[$dates[$i]]['amount'] ?? 1;
                 $previous_week = $current_week;
             }
+            $weeks[$current_week] = round($weeks[$current_week], 2);
```
jaragunde, line 191: Sorry, I cannot reproduce a situation that generates a report with long decimal numbers. We have already rounded partial leaves in `mapHalfLeaves`, which is used as input of this function. Isn't that enough?

anarute: I thought it was enough too but I think summing them in the previous for loop generates unexpected results. I could reproduce it by manually adding a holiday of 1:33hour

Screen Shot 2022-06-08 at 13 56 33

Then you can see it in the Vacation Management
Screen Shot 2022-06-08 at 13 57 53

and Vacation Summary:
Screen Shot 2022-06-08 at 13 57 43

jaragunde (approved):

Issue #511, opened 2021-09-13
Change to continous delivery model


jaragunde:
Upstream PhpReport is provided with yearly releases, but that doesn't fit known actual deployments which are updated from the repository more often. It would be ideal to align upstream with this delivery process. We would also reduce the effort done in releases, which are completely manual at this point, and minimize errors.

We would have to guarantee that  an admin is able to replace the files in their deployment with the files from the main branch in any given point (individual commits in branches merged into main would not count), and then be able to easily match their config and database. So I think we should:

* version the database (already done, maybe change from "2.x" version names to incremental integers like db_version 3, 4, etc.)
  * for every change in the DB we commit in the future, we would bump the DB and probably include an upgrade script too
* version the config file (e.g. `define(CONFIG_VERSION, 2)`)
* remove `config.php` from the repository to prevent overwriting local config (in fact, we already exclude this file from the release tarballs)
* add initial checks in the program (during login?) to find out if it's running the correct DB and config version

anarute:
@jaragunde cool, agree we should start prioritizing this.

A bit related, could you give me admin powers in the repository? I want to setup a github action to run the automated tests on pull requests and in the default branch

PR #590, opened 2022-06-09
Add updated_at field to task

anarute:
We need to know when users update their tasks, specially the ones related to their vacation plans.

This patch adds a column called `updated_at` to the task table as well as their migration scripts.

It also starts populating the field whenever a task is created or updated.

### DIFF
--- a/model/vo/TaskVO.php
+++ b/model/vo/TaskVO.php
@@ -68,9 +68,9 @@ class TaskVO {
     protected $projectId = NULL;
 
     public function setId($id) {
-    if (is_null($id))
-        $this->id = $id;
-    else
+        if (is_null($id))
+            $this->id = $id;
+        else
             $this->id = (int) $id;
     }
 
@@ -87,9 +87,9 @@ class TaskVO {
     }
 
     public function setInit($init) {
-    if (is_null($init))
-        $this->init = $init;
-    else
+        if (is_null($init))
+            $this->init = $init;
+        else
             $this->init = (int) $init;
     }
 
@@ -98,9 +98,9 @@ class TaskVO {
     }
 
     public function setEnd($_end) {
-    if (is_null($_end))
-        $this->_end = $_end;
-    else
+        if (is_null($_end))
+            $this->_end = $_end;
+        else
             $this->_end = (int) $_end;
     }
 
@@ -158,8 +158,8 @@ class TaskVO {
 
     public function setUserId($userId) {
         if (is_null($userId))
-        $this->userId = $userId;
-    else
+            $this->userId = $userId;
+        else
             $this->userId = (int) $userId;
     }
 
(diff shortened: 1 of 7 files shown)


jaragunde (commented):
There is another function in the DAO that modifies the tasks data, `PostgreSQLTaskDAO::update()`, but I'm not sure when/if we use it. We should either:

1. Set the `updated_at` field in that function too.
2. Verify whether it's actually unused and delete it!

I've been reviewing the usage of every DAO function while working on #513 to prevent rewriting unused code, so you may do only 1 this time and we leave 2 for later if you are in a hurry.

model/vo/TaskVO.php
```diff
@@ -178,6 +179,17 @@ public function getProjectId() {
         return $this->projectId;
     }
 
+    public function setUpdatedAt($updated_at) {
+        if (is_null($updated_at))
+            $this->updated_at = $updated_at;
+        else
+            $this->updated_at = (int) $updated_at;
```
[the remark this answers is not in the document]
anarute: oh jeez, bad copy/paste from my side, nice catch!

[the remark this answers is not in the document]
anarute: fixed

update/update-from-2.21-to-2.22.php
```diff
@@ -0,0 +1,58 @@
...
+
+define('PHPREPORT_ROOT', __DIR__ . '/../');
+define('SQLPATH', PHPREPORT_ROOT . 'sql/update/');
+
+/* These are the sql files that must be executed to prepare DB.
+ *
+ * IMPORTANT: they must be ordered for their proper execution.
+ */
+$sqlFiles = array();
+$sqlFiles[] = SQLPATH . "add-updated-at-to-task.sql";
```
jaragunde, line 31: I missed this before: I think we should also include `bump-db-version-2-22.sql` here. I wasn't planning to do any new "stable" releases so every change in the DB should bump its version so we can easily track where the DB is with regard to the code.

jaragunde: Related: #511.

anarute:
> There is another function in the DAO that modifies the tasks data, `PostgreSQLTaskDAO::update()`, but I'm not sure when/if we use it. We should either:
> 
> 1. Set the `updated_at` field in that function too.
> 2. Verify whether it's actually unused and delete it!
> 
> I've been reviewing the usage of every DAO function while working on #513 to prevent rewriting unused code, so you may do only 1 this time and we leave 2 for later if you are in a hurry.

I couldn't find anywhere this method was being used so I removed it.

I'm trying to port the task DAO to use PDO, but it's been a bit tricky 😬

jaragunde:
> I'm trying to port the task DAO to use PDO, but it's been a bit tricky 😬

You may want to try to migrate a small DAO first to "practice", we still have some!

update/update-from-2.21-to-2.22.php
jaragunde: I think it's all good, could you do just this small modification?

jaragunde (approved):
Thanks!

Issue #592, opened 2022-07-04
Docker image fails building


noraj:
The docker image is failing to build when running `docker build . -t phpreport`

![image](https://user-images.githubusercontent.com/16578570/177159472-103efac3-20a2-4237-831d-045ca661a645.png)

![image](https://user-images.githubusercontent.com/16578570/177159175-74fadd5b-0ea4-4cba-9625-05161806cf11.png)

as it seems the `git` command was missing it added it to the Dockerfile

By adding it the image built successfully but with some errors for some steps.

![image](https://user-images.githubusercontent.com/16578570/177160010-65e5342f-5da5-4986-a89a-fae4874cae61.png)

And when I start the container as in [the doc](https://github.com/Igalia/phpreport/blob/main/docs/admin/docker.rst), http://localhost/phpreport redirects me to http://localhost/phpreport/web/login.php but the page is just white and I get an HTTP 500 error that is probably related to all the fails during the build process.

anarute:
Thank you for your contribution @noraj!

Our Dockerfile is outdated indeed and I agree that we should use the official php image. As a first step I think it makes sense to add the docker-compose as you suggested and then improve the Dockerfile.

dmtrek14:
The Dockerfile has been revamped and should build okay now.

PR #593, opened 2022-07-04
add missing package

noraj:
git was missing, see #592


anarute:
We actually should modernize the front end build and remove the dependency of git, but it will take some time so for now let's add it, thank you!

PR #594, opened 2022-07-04
add docker compose

noraj:


jaragunde (approved):
Thanks for your contribution! This change is technically correct, but phpreport doesn't work inside that environment because we require newer versions of some base software; I'll submit a PR to address that. I'll appreciate your input!

jaragunde:
Submitted dockerfile update at PR #595. Feel free to take a look!

PR #595, opened 2022-07-21
Update docker files to use latest Ubuntu LTS.

jaragunde:
Running on the old bionic we had trouble with the PHP and UglifyJS versions, at least.


noraj (approved):
it's a start to update some components but ultimately having postgresql in another container that is launched in docker-compose would be better than using supervisord

anarute (approved):

PR #597, opened 2022-08-18
caldav: add email for calendar user

anarute:
Currently our username and email don't match so if these values
are different the calendar event won't be displayed in the user's
calendar.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -332,6 +332,7 @@ define('CALENDAR_URL', '');
 define('CALENDAR_ID', '');
 define('CALENDAR_USERNAME', '');
 define('CALENDAR_PASSWORD', '');
+define('CALENDAR_EMAIL', '');
 
 /**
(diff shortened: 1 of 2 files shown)


jaragunde:
It makes sense not building up email addresses like we did there, but does that really fix the problem you describe? What if an individual user's email and username + domain don't match?

anarute:
> It makes sense not building up email addresses like we did there, but does that really fix the problem you describe? What if an individual user's email and username + domain don't match?

That's a good point, I think the event wouldn't show up for the user. But in that case the best way to fix this in my opinion is to start storing the user email in PhpReport as well: #434

anarute:
@jaragunde gentle ping about this one :)

jaragunde (approved):
Sorry I forgot. Yeah, it's a positive change, let's land it.

PR #600, opened 2022-11-08
Add endpoint to create long periods of leaves

anarute:
This endpoint allows us to easily create long periods of leaves for users (normally needed for users that take sick leaves).

It was built in a generic way, so when making the request we need to pass the projectId as a query parameter, this way it can be used for different types of long leaves.


jaragunde (approved):
Looks good to me!

PR #602, opened 2022-11-22
Add description to the long leave form

anarute:
After showing to the stakeholders, they said it was also important to be able to fill this data.

### DIFF
--- a/web/include/form.css
+++ b/web/include/form.css
@@ -3,6 +3,16 @@
     padding: 1rem;
 }
 
+.form-title {
+    border-bottom: 1px solid #ccc;
+    padding-bottom: 10px;
+}
+
+.instructions {
+    padding: 10px 0;
+    font-size: 0.9rem;
+}
+
 .field {
     display: flex;
     margin: 10px 0;
--- a/web/include/form.css
+++ b/web/include/form.css
@@ -14,7 +14,9 @@
 }
 
 .field input,
-.field select {
+.field select,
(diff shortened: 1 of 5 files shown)


jaragunde:
Thanks, Ana, I'll take a look tomorrow. 

There's something else I've been thinking... We should probably add some documentation to the feature, specially to document potentially strange behaviors like: "I'm getting an error, a possible cause is that leave tasks are already created", because this works unlike the holiday calendar screen that removes the existing holidays before creating the new ones. We now remember how it works but we might forget in a year, or if someone wants to use it and doesn't ask.

And, given that we cannot remove created leave tasks without sysadmin implication, we might even want to add a disclaimer in the form page like "use with caution, tasks created with this form can only be removed by the user or a sysadmin".

jaragunde (approved):
It looks great, thanks!

One detail (not a blocker), you will probably want to add how to include a description to the endpoint documentation in: docs/user/users-management.rst

Even better, you could document the usage of the web form, because it doesn't have much purpose for final users to use the endpoint directly now.

anarute:
@jaragunde I've added a small warning to the form: https://github.com/Igalia/phpreport/pull/602/commits/cf77d36745de34675e2024234dcf20d2e0e38965

jaragunde (approved):
Good catch, that flat array! And the comment looks enough for a manager doing this for the first time to be extra-careful. Thank you!

PR #603, opened 2022-12-03
Allow 0 hour tasks

anarute:
Some users need to record 0-hour tasks to record when they work
or took the day off but don't want to affect the amount of worked hours.

### DIFF
--- a/model/facade/action/PartialUpdateTasksAction.php
+++ b/model/facade/action/PartialUpdateTasksAction.php
@@ -128,18 +128,26 @@ class PartialUpdateTasksAction extends Action{
                 unset($this->tasks[$i]);
             }
 
-            if ($task->isInitDirty()){
-                // Check if init was updated and endTime should be converted to 24:00
-                // (case it was a 0-hour task and it's not anymore)
+            if ($task->isInitDirty() & !$task->isEndDirty()) {
                 $currentEndTime = $oldTask->getEnd();
                 $newInitTime = $task->getInit();
+                // Check if init was updated and endTime should be converted to 24:00
+                // (in case it was a 0-hour task and it's not anymore)
                 if ($newInitTime > $currentEndTime && $currentEndTime === 0) {
                     $task->setEnd(1440);
                 }
 
                 // Check if init was updated and endTime should be converted to 0h
-                // (case it was a regular task and now it's a 0-hour task)
+                // (in case it was a regular task and now it's a 0-hour task)
                 if ($newInitTime === 0 && $currentEndTime == 1440) {
+                    $task->setEnd(0);
+                }
+            } elseif ($task->isEndDirty()) {
+                $currentInitTime = $oldTask->getInit();
+                $newEndTime = $task->getEnd();
+                // Check if end was updated and it should be converted to 0
+                // (in case it became a 0-hour task)
+                if ($currentInitTime === 0 && $newEndTime === 1440) {
                     $task->setEnd(0);
                 }
             }
--- a/model/facade/action/PartialUpdateTasksAction.php
+++ b/model/facade/action/PartialUpdateTasksAction.php
@@ -127,6 +127,22 @@ class PartialUpdateTasksAction extends Action{
                 $discardedTasks[] = $task;
                 unset($this->tasks[$i]);
             }
+
+            if ($task->isInitDirty()){
+                // Check if init was updated and endTime should be converted to 24:00
+                // (case it was a 0-hour task and it's not anymore)
+                $currentEndTime = $oldTask->getEnd();
+                $newInitTime = $task->getInit();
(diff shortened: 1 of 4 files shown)


eocanha:
Thanks a lot for working on this!

jaragunde (commented):
I found an issue:

1. Create an empty task with times 0:00 - 0:00
2. Update init time to 1:00

At this point, save will fail with this error:
```
[Mon Dec 12 13:02:20 2022] [::1]:54134 [500]: POST /web/services/updateTasksService.php - Uncaught SQLQueryErrorException: ERROR:  23514: new row for relation "task" violates check constraint "end_after_init_task"
DETAIL:  Failing row contains (33, 2022-12-12, 60, 0, null, f, null, null, null, 2, 2, null, null, f, 2022-12-12 13:02:20.070846).
SCHEMA NAME:  public
TABLE NAME:  task
CONSTRAINT NAME:  end_after_init_task
LOCATION:  ExecConstraints, execMain.c:1934 in /var/www/html/phpreport/model/dao/TaskDAO/PostgreSQLTaskDAO.php:706
Stack trace:
#0 /var/www/html/phpreport/model/dao/TaskDAO/PostgreSQLTaskDAO.php(735): PostgreSQLTaskDAO->partialUpdate()
#1 /var/www/html/phpreport/model/facade/action/PartialUpdateTasksAction.php(132): PostgreSQLTaskDAO->batchPartialUpdate()
#2 /var/www/html/phpreport/model/facade/action/Action.php(112): PartialUpdateTasksAction->doExecute()
#3 /var/www/html/phpreport/model/facade/TasksFacade.php(154): Action->execute()
#4 /var/www/html/phpreport/web/services/updateTasksService.php(260): TasksFacade::PartialUpdateReports()
#5 {main}
  thrown in /var/www/html/phpreport/model/dao/TaskDAO/PostgreSQLTaskDAO.php on line 706

File: model/facade/action/PartialUpdateTasksAction.php [hunk 1]
```diff
@@ -127,6 +127,20 @@ protected function doExecute() {
...
+            $newInitTime = $this->tasks[$i]->getInit();
+            if ($newInitTime > $currentEndTime && $currentEndTime === 0) {
+                $this->tasks[$i]->setEnd(1440);
+            }
+
+            // Check if init was updated and endTime should be converted to 0h
+            // (case it was a regular task and now it's a 0-hour task)
+            if ($newInitTime === 0 && $currentEndTime == 1440) {
+                $this->tasks[$i]->setEnd(0);
+            }
```
jaragunde, line 143: It's possible to call the update operation with no init time set, meaning that you don't want to update that field.

In that case, `$newInitTime` would be null. It looks like the code works properly because null is not bigger than $currentEndDate nor it equallls (===) zero, but I would wrap all this block in `if $task->isInitTimeDirty() ...`. I would find it easier to understand what's going on and prevent exotic comparisons.

anarute: That makes sense, I've just updated the last commit with this suggestion

model/facade/action/PartialUpdateTasksAction.php
```diff
@@ -127,6 +127,20 @@ protected function doExecute() {
                 $discardedTasks[] = $task;
                 unset($this->tasks[$i]);
             }
+
+            // Check if init was updated and endTime should be converted to 24:00
+            // (case it was a 0-hour task and it's not anymore)
+            $currentEndTime = $oldTask->getEnd();
+            $newInitTime = $this->tasks[$i]->getInit();
```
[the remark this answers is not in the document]
anarute: done, thank you

model/facade/action/PartialUpdateTasksAction.php
```diff
@@ -127,6 +127,22 @@ protected function doExecute() {
...
+                $newInitTime = $task->getInit();
+                if ($newInitTime > $currentEndTime && $currentEndTime === 0) {
+                    $task->setEnd(1440);
+                }
+
+                // Check if init was updated and endTime should be converted to 0h
+                // (case it was a regular task and now it's a 0-hour task)
+                if ($newInitTime === 0 && $currentEndTime == 1440) {
+                    $task->setEnd(0);
+                }
```
jaragunde, line 144: Sorry, a side effect of my last request is that this block doesn't run when you update *the end time only*. So, if you create a task from 0:00 to 1:00, save, then update end time to 0:00, it will still be recorded as a 24:00 task.

anarute: I think I've covered it now, hopefully I covered all scenarios 😅

anarute:
@jaragunde gentle ping about this one :)

jaragunde (approved):
Sorry for the delay. This looks good to me now, I didn't manage to break it  :wink: . Thanks!

PR #604, opened 2022-12-14
Add year filter in vacation summary

anarute:
A very quick hack to allow users to see next year's summary, this patch allows filtering the report by passing the desired year in the url, for example: `/web/holidaySummary.php?year=2023`

Next step would be adding a component in the UI to allow them to do it more easily.

### DIFF
--- a/web/holidaySummary.php
+++ b/web/holidaySummary.php
@@ -52,6 +52,9 @@ include_once("include/header.php");
                 </div>
             </div>
         </div>
+        <div class="content">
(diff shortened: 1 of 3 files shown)


jaragunde (approved):
I took a quick look and looks good, I understand we have some hurry to land this so I'm just approving it :)

anarute:
> I took a quick look and looks good, I understand we have some hurry to land this so I'm just approving it :)

The hurry is more related to end of the year planning, if we land it by January it loses the usefulness a bit hehe thank you!

PR #607, opened 2023-01-12
Fix Vacation Summary report Used Hours column

anarute:
In case the user was filtering the report for another year that was not the current one, the Used Hours column would be wrong because it was filtering the hours until the current day instead of the last day of the report year. This would result in a range of dates that spawns from Jan 1st of the filtered year until the current day.

We want to use the current date as end date only if we are filtering for the current year.

PS: I really tried to create tests for this but the dependencies are too coupled and would required a big refactoring that I would rather not do it now - it could be lack of skills from my side as well but from what I digged into PhpUnit and how it works, mocking static methods and classes called inside other classes is very painful, ideally we want to decouple the dependencies first, like I did for [LoginManager here](https://github.com/Igalia/phpreport/blob/main/web/services/HolidayService.php#L36).

### DIFF
--- a/model/facade/action/GetHolidaySummaryReportAction.php
+++ b/model/facade/action/GetHolidaySummaryReportAction.php
@@ -53,12 +53,19 @@ class GetHolidaySummaryReportAction extends GetHolidayHoursBaseAction
             $this->user,
             $this->end
         );
-        $summaryForToday = \UsersFacade::GetHolidayHoursSummary(
-            $this->init,
-            $this->end,
-            $this->user,
-            $today
-        );
+        $usedHours = $summary['usedHours'][$this->user->getLogin()];
+        if (date('Y') == $this->end->format('Y')) {
+            // If today's year the same as the one being filtered, we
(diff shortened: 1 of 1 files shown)


dmtrek14 (approved):
@anarute This change looks good to me.

jaragunde:
Looks good to me, too! Thanks!

Regarding testing, the most achievable path I see is to create a testing DB and having tests use the service endpoints. It would be something simple, query an endpoint with some known parameters and compare the output with some text-file expectations of the output before the change. That way we would test a "vertical slice" of our code, and a very meaningful one.

I don't know how we could setup that in a CI environment, but I think we should aim for this instead of writing a lot of code to support tests.

jaragunde (approved):

jaragunde:
> Yep, that's what I was concluding too.. I have some ideas on how to do it, I'll start working on it soon.

That sounds great! Even if we had the most basic setup to call the endpoints, generate expectations and compare, that would be awesome because we could run it on a set of test or production data as part of our development process!

Issue #496, opened 2021-04-23
Do something meaningful with SQL exceptions


jaragunde:
Certain DAO operations may throw SQL exceptions, one example is `PostgreSQLTaskDAO::create()`, which may throw `SQLUniqueViolationException` or `SQLQueryErrorException`. We currently let them flow up in the code, eventually crashing and resulting in an internal server error and a HTTP status 500 in the response.

We should make something with these exceptions and make the backend fail gracefully, returning a meaningful error message.

This is the cause of some reported errors, like #492 and #495. It's very related with #173, too.

jaragunde:
This is also the root cause behind #110.

jaragunde:
We don't have a good way in the business logic layer (facades & actions) to pass detailed information to the upper layer in case of error. I would create an OperationResult class that would be returned by CRUD operations in the business logic layer. This class would have a boolean success field, and fields for an error number and message. It could even contain a list of tasks, the ones that we failed to create/update/delete.

Then, web services would check the OperationResult and return the information in it in case of error.

jaragunde:
Danielle started working on this problem and I've been following up. Some fixes have been merged lately.

From PR #608:
```
a968bd17 Merge pull request #608 from Igalia/bubble-up-sql-errors [Danielle M]
7ec2b2fe Fix string construction in service [Danielle Mayabb]
207ea879 (origin/bubble-up-sql-errors) Clean up formatting, use getters [Danielle Mayabb]
5bfa360e Clean up some xml so that messages are returned correctly [Danielle Mayabb]
a4175f5b Fix formatting/naming, return error code on error [Danielle Mayabb]
cfcdc447 Reload data store if creation fails [Danielle Mayabb]
a15d8cae Update save of dates, facade, and service [Danielle Mayabb]
d1940ea5 Add error handling for not null violation [Danielle Mayabb]
7762bffd Update object metadata and fix date conversion [Danielle Mayabb]
b9d4be4a Update copyright and remove trailing whitespace [Danielle Mayabb]
6c080269 Update var convention and delete whitespace [Danielle Mayabb]
90337f62 Expand exception handling in projectManagement.js [Danielle Mayabb]
6ad9871c Update createProjectsService to check new property isSuccessful [Danielle Mayabb]
074632f4 Update Project action, facade, and DAO [Danielle Mayabb]
d58e254f Update ProjectDAO create to use PDO and return OperationResult [Danielle Mayabb]
[...]

PR #608, opened 2023-01-13
Add OperationResult class and begin handling sql errors more gracefully

dmtrek14:
This PR begins to address #496 by adding an OperationResult class and updates code related to the creation of Projects to begin using this class. Changes in these commits include:
- Add the OperationResult class
- Refactor Project DAO and Facade to use this class
- Update createProjectsService to check class
- Update projectManagement.js exception handler for project creation to pass along the message from the back end.

Comments and suggestions welcome as my PHP is a bit rusty.

### DIFF
--- a/model/OperationResult.php
+++ b/model/OperationResult.php
@@ -32,51 +32,52 @@
  *
  *  This is the object returned for all CRUD operations
  *
-  *  @property boolean $isSuccessful Whether the operation was successful or not
-  *  @property int $responseCode The http status code
-  *  @property int $errorNumber The php error number
-  *  @property string $message Error message that is passed up to front end. Can be php error or custom string.
+ *  @property boolean $isSuccessful Whether the operation was successful or not
+ *  @property int $responseCode The http status code
+ *  @property int $errorNumber The php error number
+ *  @property string $message Error message that is passed up to front end. Can be php error or custom string.
  */
-class OperationResult {
-	public $isSuccessful;
-	public $responseCode;
-	public $errorNumber;
-	public $message;
+class OperationResult
+{
+    private $isSuccessful;
+    private $responseCode;
+    private $errorNumber;
+    private $message;
 
-	function __construct($isSuccessful, $responseCode = null) {
-		$this->isSuccessful = $isSuccessful;
-		$this->responseCode = $responseCode;
-	}
+    function __construct($isSuccessful, $responseCode = null) {
+        $this->isSuccessful = $isSuccessful;
+        $this->responseCode = $responseCode;
+    }
 
-	public function getIsSuccessful(){
-		return $this->isSuccessful;
-	}
+    public function getIsSuccessful(){
+        return $this->isSuccessful;
+    }
 
-	public function setIsSuccessful($isSuccessful){
-		$this->isSuccessful = (boolean) $isSuccessful;
-	}
+    public function setIsSuccessful($isSuccessful){
+        $this->isSuccessful = (boolean) $isSuccessful;
+    }
 
-	public function getResponseCode(){
-		return $this->responseCode;
-	}
+    public function getResponseCode(){
+        return $this->responseCode;
+    }
 
-	public function setResponseCode($responseCode){
-		$this->responseCode = (int) $responseCode;
-	}
+    public function setResponseCode($responseCode){
+        $this->responseCode = (int) $responseCode;
+    }
 
-	public function getErrorNumber(){
-		return $this->errorNumber;
-	}
+    public function getErrorNumber(){
+        return $this->errorNumber;
+    }
 
-	public function setErrorNumber($errorNumber){
-		$this->errorNumber = (int) $errorNumber;
-	}
+    public function setErrorNumber($errorNumber){
+        $this->errorNumber = (int) $errorNumber;
+    }
 
-	public function getMessage(){
-		return $this->message;
-	}
+    public function getMessage(){
+        return $this->message;
+    }
 
-	public function setMessage($message){
-		$this->message = (string) $message;
-	}
+    public function setMessage($message){
+        $this->message = (string) $message;
+    }
 }
--- a/model/OperationResult.php
+++ b/model/OperationResult.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright (C) 2009 Igalia, S.L. <info@igalia.com>
+ * Copyright (C) 2023 Igalia, S.L. <info@igalia.com>
  *
  * This file is part of PhpReport.
  *
@@ -33,7 +33,7 @@
  *  This is the object returned for all CRUD operations
  *
   *  @property boolean $isSuccessful Whether the operation was successful or not
-  *  @property int $responseCode The http status code 
+  *  @property int $responseCode The http status code
   *  @property int $errorNumber The php error number
   *  @property string $message Error message that is passed up to front end. Can be php error or custom string.
  */
--- a/model/OperationResult.php
+++ b/model/OperationResult.php
@@ -1,0 +1,82 @@
+<?php
+/*
+ * Copyright (C) 2009 Igalia, S.L. <info@igalia.com>
+ *
+ * This file is part of PhpReport.
+ *
+ * PhpReport is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PhpReport is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PhpReport.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+/** File for OperationResult
+ *
+ *  This file just contains {@link OperationResult}.
+ *
+ * @filesource
+ * @package PhpReport
+ * @author Danielle Mayabb <danielle@igalia.com>
+ */
+
+/** Result returned after CRUD operations
+ *
+ *  This is the object returned for all CRUD operations
+ *
+  *  @property boolean $isSuccessful Whether the operation was successful or not
+  *  @property int $responseCode The http status code 
+  *  @property int $errorNumber The php error number
+  *  @property string $message Error message that is passed up to front end. Can be php error or custom string.
+ */
+class OperationResult {
+	public $isSuccessful;
+	public $responseCode;
+	public $errorNumber;
+	public $message;
+
+	function __construct($isSuccessful, $responseCode = null) {
+		$this->isSuccessful = $isSuccessful;
+		$this->responseCode = $responseCode;
+	}
+
+	public function getIsSuccessful(){
+		return $this->isSuccessful;
+	}
+
+	public function setIsSuccessful($isSuccessful){
+		$this->isSuccessful = (boolean) $isSuccessful;
+	}
+
+	public function getResponseCode(){
+		return $this->responseCode;
+	}
+
+	public function setResponseCode($responseCode){
+		$this->responseCode = (int) $responseCode;
+	}
+
+	public function getErrorNumber(){
+		return $this->errorNumber;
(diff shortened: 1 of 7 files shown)


web/js/projectManagement.js
```diff
@@ -1100,18 +1100,21 @@ Ext.onReady(function(){
...
                 projectGrid.getSelectionModel().selectRow(rowIndex);
                 projectGrid.getView().focusRow(rowIndex);
             },
-            'exception': function(){
-                App.setAlert(false, "Some Error Occurred While Saving The Changes");
-            },
+            'exception': function(proxy, type, action, exception, res){
+                var parser = new DOMParser();
+                var errorDoc = parser.parseFromString(res.responseText, "text/xml");
+                var errorMessage = errorDoc.getElementsByTagName("error")[0].childNodes[0].nodeValue;
```
anarute, line 1112: I'm not sure how the format of the XML response is but is it possible to have multiple error elements? In this case we may need to be more specific to be sure we are displaying all the errors, but I'm happy to fetch this and test it next week as well as I'm not completely sure how it works.

anarute (commented):
That looks pretty good! I've added some comments but they are all details, on Monday I can run it locally and help testing too.

web/js/projectManagement.js
```diff
@@ -1100,18 +1100,21 @@ Ext.onReady(function(){
         listeners: {
             'write': function (store, action, result, res, rs) {
                 App.setAlert(true, "Projects Changes Saved");
-
+
```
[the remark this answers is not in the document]
dmtrek14: Eek. Sorry. I normally have something like that, but don't think I got it set in vscodium. Gonna do that now bc extra space bothers me, too.

web/js/projectManagement.js
```diff
@@ -1100,18 +1100,21 @@ Ext.onReady(function(){
...
                 // select newly added row and scroll to it
                 var rowIndex = store.indexOf(rs);
                 projectGrid.getSelectionModel().selectRow(rowIndex);
                 projectGrid.getView().focusRow(rowIndex);
             },
-            'exception': function(){
-                App.setAlert(false, "Some Error Occurred While Saving The Changes");
-            },
+            'exception': function(proxy, type, action, exception, res){
+                var parser = new DOMParser();
```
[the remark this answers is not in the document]
dmtrek14: Perfect. I cringed a little when matching the code style of older JS.

web/js/projectManagement.js
dmtrek14: We should be able to. Since we are sending back xml to the front end, I think we could add as many <error> items to the object as needed and then just parse them out to the alert. I'll test the js here with multiple <error> elements; I probably need to modify the last bit of code since it's taking the first items out of the arrays instead of looping.

jaragunde (commented):
Thanks a lot for working on this! I always feel sad whenever I get one of those indescriptive error messages. It's also a great way to get a feel of the project, working across all the layers.

model/dao/ProjectDAO/PostgreSQLProjectDAO.php
```diff
@@ -594,36 +595,58 @@ public function update(ProjectVO $projectVO) {
...
+            $result->setResponseCode(201);
+        }
+        catch (PDOException $ex) {
+            //make sure to log the error as a failure, but return the OperationResult object
+            //successfully so that front end can see error and fail gracefully
+            $errorMessage = $ex->getMessage();
+            error_log('Project creation failed: ' . $errorMessage);
+            $result->setErrorNumber($ex->getCode());
+            $resultMessage = "Project creation failed: \n";
+            if(strpos($errorMessage, "Foreign key violation")){
```
jaragunde, line 636: A "not null" error is also possible, e.g. when you forget to set the area:
```
[Mon Jan 16 10:13:02 2023] Project creation failed: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "areaid" of relation "project" violates not-null constraint
```

jaragunde: Null customer is technically allowed although the UI tries to prevent it, we made a mess with that :shrug: :sweat_smile:

jaragunde:
Maybe you already know or have a better way to do it, but there's a `web/APITest.php` page where you can send any random request to make the endpoint fail in the ways you want.

model/OperationResult.php
```diff
@@ -0,0 +1,82 @@
...
+/** Result returned after CRUD operations
+ *
+ *  This is the object returned for all CRUD operations
+ *
+  *  @property boolean $isSuccessful Whether the operation was successful or not
+  *  @property int $responseCode The http status code 
+  *  @property int $errorNumber The php error number
+  *  @property string $message Error message that is passed up to front end. Can be php error or custom string.
+ */
+class OperationResult {
```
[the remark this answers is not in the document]
jaragunde: The VO objects represent an entity in the domain, so I think this class doesn't fit there.

model/facade/ProjectsFacade.php
```diff
@@ -147,17 +147,13 @@ static function CreateProject(ProjectVO $project) {
...
     static function CreateProjects($projects) {
 
-    foreach((array)$projects as $project)
-        if ((ProjectsFacade::CreateProject($project)) == -1)
-            return -1;
-
-    return 0;
-
+        foreach ((array) $projects as $project)
+            return ProjectsFacade::CreateProject($project);
```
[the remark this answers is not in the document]
dmtrek14: I didn't see anywhere that we create multiple projects, either. Do either of you know of a use case from other teams where this functionality would be desired (even if not implemented at this time)?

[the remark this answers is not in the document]
dmtrek14: I chatted a little with @anarute on this and we decided that we'd leave the ability for creation of multiple projects for the moment. I have made a couple of small code changes that would better handle multiple errors in this case - I updated the method in the ProjectsFacade to return an array of OperationResults and some code in the service for building the return xml.

dmtrek14:
@anarute That's a good catch. I just added a call to reload the data store if creation fails so that we don't have a bad phantom record listed.

web/services/createProjectsService.php
```diff
@@ -208,11 +209,37 @@
...
+            $operationResult = ProjectsFacade::CreateProject($createProjects[0]);
+            if (!$operationResult->isSuccessful)
+                $string = "<return service='createProjects'><error id='1'>". $operationResult->message . "</error></return>";
+        }
+        else {
+            //leaving the possibility to create multiple projects in the future
+            $operationResults = ProjectsFacade::CreateProjects($createProjects);
+            $string = "<return service='createProjects'>";
+            $errors = array_filter($operationResults, function ($item) {
+                return ($item->isSuccessful = false);
```
jaragunde, line 222: It should be `==`

jaragunde (commented):
Looks good to me! I just caught a small mistake, see below. Thanks!

web/services/createProjectsService.php
```diff
@@ -208,11 +209,37 @@
...
+            });
+            if($errors){
```
jaragunde, line 224: Indentation is weird in the block below.

web/services/createProjectsService.php
```diff
@@ -208,11 +209,37 @@
...
+            $string .= "<errors>";
+                foreach((array) $errors as $result){
+                    if (!$result->isSuccessful)
+                        $string .= "<error id=''>" . $result->message . "</error>";
+                }
+            $string .= "</errors>";
+
+            }
+            else {
+            $string .= "<ok>All operations successful</ok>";
```
jaragunde, line 234: I think you should let code flow into the `if (!isset($string))` block below in case of success. One of the things that block does is to return the data of the created projects including their newly assigned IDs, which the front-end can use to issue update operations later.

It's unclear what to do if some projects fail and some don't, but we'll figure out if we ever use the feature...

web/services/createProjectsService.php
```diff
@@ -208,11 +209,37 @@
 
 
     if (count($createProjects) >= 1)
-        if (ProjectsFacade::CreateProjects($createProjects) == -1)
-            $string = "<return service='createProjects'><error id='1'>There was some error while creating the projects</error></return>";
+        if(count($createProjects) == 1){
+            $operationResult = ProjectsFacade::CreateProject($createProjects[0]);
+            if (!$operationResult->isSuccessful)
+                $string = "<return service='createProjects'><error id='1'>". $operationResult->message . "</error></return>";
```
jaragunde, line 215: We still return code 200 in case of error, we should probably change it with `http_response_code()`. Not your fault, this service didn't do it, but newer code by Ana does :slightly_smiling_face:

jaragunde (commented):
My mind drifted back to the review and I added a few more comments... :innocent:

web/services/createProjectsService.php
anarute: Here's an example https://github.com/Igalia/phpreport/blob/main/web/services/getHolidays.php#L43

web/services/createProjectsService.php
jaragunde: This is still not working as expected. The service returns this in case of success creating one project:
```
<?xml version="1.0"?>
<return service="createProjects"><error id="1">Project created successfully.</error></return>
```

And this in case of multiple projects:
```
<?xml version="1.0"?>
<return service="createProjects"/>
```

dmtrek14: Looks like my xml was a bit malformed (and formed at the wrong time).

model/OperationResult.php
```diff
@@ -0,0 +1,82 @@
...
+	public $responseCode;
+	public $errorNumber;
+	public $message;
+
+	function __construct($isSuccessful, $responseCode = null) {
+		$this->isSuccessful = $isSuccessful;
+		$this->responseCode = $responseCode;
+	}
+
+	public function getIsSuccessful(){
```
jaragunde, line 51: Sorry I didn't review carefully before, I just noticed this file is indented with tabs.

Also, if we have a getter we should use it, and have the attributes private.

web/services/createProjectsService.php
```diff
@@ -208,11 +209,33 @@
 
 
     if (count($createProjects) >= 1)
-        if (ProjectsFacade::CreateProjects($createProjects) == -1)
-            $string = "<return service='createProjects'><error id='1'>There was some error while creating the projects</error></return>";
+        if(count($createProjects) == 1){
+            $operationResult = ProjectsFacade::CreateProject($createProjects[0]);
+            if ($operationResult->isSuccessful == false){
```
jaragunde, line 214: It's so weird to have to do this... Maybe because `isSuccessful` was never initialized in some cases? We could initialize it to false in the `OperationResult`.

Also, I'd use the getter and not the property directly.

jaragunde (commented):
Sorry I noticed a few more issues, but this is looking great. Thanks a lot!

model/OperationResult.php
anarute: We should setup a linter and automated checks soon :)

dmtrek14: @jaragunde No problem! This is how I learn to write better PHP ;-)

@anarute A linter is a great idea. It would save me from myself.

web/services/createProjectsService.php
dmtrek14: Sounds good to me. Will make the changes and test it out.

web/services/createProjectsService.php
```diff
@@ -208,11 +209,34 @@
...
+        else {
+            //leaving the possibility to create multiple projects in the future
+            $operationResults = ProjectsFacade::CreateProjects($createProjects);
+            $errors = array_filter($operationResults, function ($item) {
+                return (!$item->getIsSuccessful());
+            });
+            if($errors){
+                //if multiple failures, let's just return a 500
+                http_response_code(500);
+                $string .= "<return service='createProjects'><errors>";
```
jaragunde, line 228: This produces an "undefined variable" error, because it's the first time you use `$string` in this code path. Just use the `=` operator instead of `.=` :slightly_smiling_face:

jaragunde (approved):
I caught a PHP warning but, other than that, my tests were successful. Thanks!

web/services/createProjectsService.php
dmtrek14: Ah! Good catch. Will definitely fix!

anarute (approved):

PR #610, opened 2023-01-24
Add initial test environment

anarute:
Create a docker compose that spins up 2 containers, one for the app and another for the database that will already be filled with initial data that will be used for testing.

These containers will be used for automated tests, that's why the php application is run with it's built-in web server instead of a more robust solution and also why the javascript files are not minified.


anarute:
@dmtrek14 thanks for the input! It all makes sense, so I'll rename the files and move them to a docker folder already.

jaragunde (commented):
I don't have an environment set up for this, but it looks good on a first read. Good job!

dmtrek14:
This looks good to me!

dmtrek14 (approved):

PR #612, opened 2023-02-07
611 midnight task error

### LINKED ISSUE #611: Error 500 when creating new task ending at 00:00 if start time is set later
1. Go to a day with no task, or click to add a new task, so that all the fields are empty
2. Set the end time to `00:00`
3. Set the start time, e.g. `23:00`
4. Choose some project
5. Save

Expected: it works.  
Actual: "Some Error Occurred While Saving The Changes (please check you haven't clipped working hours)".

The server is producing an error 500. I think it's because

https://github.com/Igalia/phpreport/blob/78a4400a77aa1b3cf7150d7f06907ada2291f758/web/services/createTasksService.php#L126

where `$initTime` is an undefined variable. Maybe set `$initTime = NAN` at the top, or check `isset($initTime)`.

dmtrek14:
Fixes #611 

I removed the check against `$initTime` out of the parser switch statement for the `$endTime` block since values are set in the order the user set them, which could cause errors due to variables/values not being set yet. I instead make the check after the switch block is done.

### DIFF
--- a/web/services/createTasksService.php
+++ b/web/services/createTasksService.php
@@ -217,7 +217,6 @@
                 }
                 //Support 0-hour tasks: reparse end time if initTime == 0 to the end so that order of parse doesn't cause error if end time added before init time by users
                 if (($endTimeParseOrig['hour']==0) && ($endTimeParseOrig['minute']==0) && ($initTime == 0)) {
-                    $endTimeParse = date_parse_from_format($endTimeFormat, $endTime);
                     $endTime = $endTimeParseOrig['hour']*60 + $endTimeParseOrig['minute'];
                     $taskVO->setEnd($endTime);
                 }
--- a/web/services/createTasksService.php
+++ b/web/services/createTasksService.php
@@ -217,7 +217,7 @@
                 }
                 //Support 0-hour tasks: reparse end time if initTime == 0 to the end so that order of parse doesn't cause error if end time added before init time by users
                 if (($endTimeParseOrig['hour']==0) && ($endTimeParseOrig['minute']==0) && ($initTime == 0)) {
-                    $endTime = $endTimeParseOrig['hour']*60 + $endTimeParseOrig['minute'];
+                    $endTime = 0;
                     $taskVO->setEnd($endTime);
                 }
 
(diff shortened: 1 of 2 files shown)


web/services/createTasksService.php
```diff
@@ -213,6 +215,11 @@
                     $string = "<return service='createTasks'><success>false</success><error id='4'>projectId is not valid</error></return>";
                     break;
                 }
+                //Support 0-hour tasks: reparse end time if initTime == 0 to the end so that order of parse doesn't cause error if end time added before init time by users
+                if (($endTimeParseOrig['hour']==0) && ($endTimeParseOrig['minute']==0) && ($initTime == 0)) {
+                    $endTime = $endTimeParseOrig['hour']*60 + $endTimeParseOrig['minute'];
```
jaragunde, line 220: Isn't `$endTime` always 0 in this block, because `($endTimeParseOrig['hour']==0) && ($endTimeParseOrig['minute']==0)`?

anarute: True, it will always be 0 in this case so we can hardcorde it directly

```suggestion
                    $endTime = 0;
```

dmtrek14: Good catch!

jaragunde (approved):
Looks good, thanks for your work!

anarute:
There is an odd corner case that still happens when you create a task that ends at 00:00 and fill the end time first, save, then changes the init time without reloading the page:

[Screencast from 2023-02-08 16-45-46.webm](https://user-images.githubusercontent.com/333447/217579745-0066e78e-9e69-47a6-aed8-dcf249a5f118.webm)

jaragunde:
Oh, good catch Ana!

dmtrek14:
Good catch, indeed. I copied code over to the updateTasksService that makes the same check and I think this solves the issue.

anarute (approved):
That makes sense, looking good now, thanks!

For this PR I would squash the commits, so we have only one with the whole change, in case we need to debug in the future it will be easier :)

dmtrek14:
I agree. A squash makes good sense in this case.

PR #613, opened 2023-02-10
Check for existence of variables before doing anything else

dmtrek14:
In fixing #611, I inadvertantly created the same issue in the updateTasksService that was originally happening in the createTasksService - variables not initialized before code tries to check them. in this instance, if someone updates any property(ies) that are **not** the endTime or initTime, an error will be thrown. This is a simple check to first make sure we have the variables we need before doing further processing. 

I could do the check all in one statement instead of having the 2 "if" blocks, if it is preferred to be more concise.


anarute (approved):
Thank you for the quick fix

jaragunde:
Indeed, thanks!

PR #615, opened 2023-02-23
Improvements to project CRUD operations

jaragunde:
I've added some fixes for the create operation, but this is mostly about the update operation:
* Manage SQL exceptions and return them as errors in the web service
* Port operation to PDO

Some things I'd like to comment:
* The end result of the update operation looks pretty terrible, with two ifs per field, but I didn't find a way to make it better.
* I realized the update operation allows to update only a subset of the project fields but the UI sends all the fields, regardless of them being modified or not. There is a separate, unused DAO operation to update a project in full. Is it worth keeping both?
* As a consequence of the above, there were many "undefined array key" warnings that were hidden in the DAO.
* When we implemented the update operation for tasks, we created `DirtyTaskVO` to keep track of what needs update, but we followed a different approach in the project update operation. I'm not changing that (yet?).


web/services/updateProjectsService.php
```diff
@@ -66,7 +66,19 @@
...
+            "end"         => false,
+            "invoice"     => false,
+            "estHours"    => false,
+            "description" => false,
+            "areaId"      => false,
+            "customerId"  => false,
+            "type"        => false,
+            "movedHours"  => false,
+            "schedType"   => false
+        );
```
jaragunde, line 81: Prevents "undefined array key" warnings down in the DAO.

dmtrek14 (approved):
This looks good to me. I tested adding and updating several projects and don't run into any unexpected regressions. 

I like the idea of simplifying code whenever it's feasible, so I'm open to removing code we don't use. For updates, I've always preferred having the whole object sent along through the stack and updating the whole object over parsing through to see what has been changed. It feels more straight-forward and is perhaps less work overall. But I know there are many places in phpreport where we are not sending the whole item through the update service, so I do see the value here of doing the work to parse what has changed and doing the partial update since doing a blanket update could force us to lose data or throw errors. So, if this means getting rid of the update method on the whole object in favor of always doing a partial update, I'm open.

jaragunde:
I think we don't have a consistent criterion about updates in the application, some objects can be partially updated but for some (most?) we send the whole object. It makes sense to rethink it, even remove the ability to partially update projects that the UI is not actually using.

jaragunde:
Hi Danielle, I was in the mood to remove code, so I went for removing *the partial update operation* for projects, which means the most code removed :laughing: 

We don't use the ability to partially update projects in the web UI, there aren't any external tools doing it either, and for any other objects in the DB we do full updates except for Tasks (which have a more robust implementation, anyway).

The latest commit does this:

    We think the ability of updating only some fields of the project is
    not important, and we prefer to minimize the amount of code to
    maintain. So we have:
    * Switched the web service to the previously unused UpdateProjectAction
      and ProjectDAO::update.
    * Modify ProjectDAO::update to detect specific errors and return an
      OperationResult, just like the partialUpdate operation did.
    * Modify ProjectDAO::update to use PDO.
    * Remove all code related to PartialUpdateProject.

anarute:
> For updates, I've always preferred having the whole object sent along through the stack and updating the whole object over parsing through to see what has been changed.

I second this, once I spent a considerable amount of time debugging something to finally understand that the front end was passing only the fields modified by the user. I agree to make this the default behavior across the project.

anarute (approved):
It looks great, thank you!

PR #618, opened 2023-02-28
Report detailed error messages on task update and delete

jaragunde:
I'm increasing the error reporting for task update and delete operations, and also making some changes in the front-end to make error messages more noticeable; I'm afraid that users have become accustomed to ignore them.

### DIFF
--- a/model/dao/TaskDAO/PostgreSQLTaskDAO.php
+++ b/model/dao/TaskDAO/PostgreSQLTaskDAO.php
@@ -707,13 +707,14 @@ class PostgreSQLTaskDAO extends TaskDAO{
             $resultMessage = "Task update failed:\n";
             if(strpos($errorMessage, "end_after_init_task")) {
                 $resultMessage .= "Start time later than end time.";
+                $result->setResponseCode(400);
             }
             else {
                 $resultMessage .= $errorMessage;
+                $result->setResponseCode(500);
             }
             $result->setMessage($resultMessage);
             $result->setIsSuccessful(false);
-            $result->setResponseCode(500);
         }
         else if (pg_affected_rows($res) == 1) {
             $result->setIsSuccessful(true);
@@ -742,7 +743,7 @@ class PostgreSQLTaskDAO extends TaskDAO{
             $result = new OperationResult(false);
             $result->setErrorNumber(10);
             $result->setMessage("Task update failed:\nDetected overlapping times.");
-            $result->setResponseCode(500);
+            $result->setResponseCode(400);
             return array($result);
         }
 
@@ -910,6 +911,7 @@ class PostgreSQLTaskDAO extends TaskDAO{
             $resultMessage = "Error creating task:\n";
             if(strpos($errorMessage, "end_after_init_task")) {
                 $resultMessage .= "Start time later than end time.";
+                $result->setResponseCode(400);
             }
             else if(strpos($errorMessage, "Not null violation")) {
                 $resultMessage .= "Missing compulsory data";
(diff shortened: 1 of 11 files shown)


jaragunde:
I've just fixed some copy-paste in error messages but this is ready to review, I'm not working more on it.

model/dao/TaskDAO/PostgreSQLTaskDAO.php
```diff
@@ -703,11 +702,27 @@ public function partialUpdate(DirtyTaskVO $taskVO) {
...
+            $resultMessage = "Task update failed:\n";
+            if(strpos($errorMessage, "end_after_init_task")) {
+                $resultMessage .= "Start time later than end time.";
+            }
+            else {
+                $resultMessage .= $errorMessage;
+            }
+            $result->setMessage($resultMessage);
+            $result->setIsSuccessful(false);
+            $result->setResponseCode(500);
```
anarute, line 716: I know we've been using 500 everywhere but I wonder if it shouldn't be 400 in this case, what do you think? Here and in the other places where the request fails because the user sent a bad input.

model/facade/action/PartialUpdateTasksAction.php
```diff
@@ -153,17 +189,13 @@ protected function doExecute() {
             }
         }
 
-        if ($taskDao->batchPartialUpdate($this->tasks) < count($this->tasks)) {
-            return -1;
-        }
+        $results = $taskDao->batchPartialUpdate($this->tasks);
 
         //TODO: do something meaningful with the list of discarded tasks
```
anarute, line 194: I think with your changes we can remove this comment? :)

model/dao/TaskDAO/PostgreSQLTaskDAO.php
```diff
@@ -952,27 +952,33 @@ public function batchCreate($tasks) {
...
+        try {
+            $statement = $this->pdo->prepare($sql);
+            $statement->bindValue(":id", $taskVO->getId(), PDO::PARAM_INT);
+            $statement->execute();
 
-            $res = pg_query($this->connect, $sql);
-            if ($res == NULL) throw new SQLQueryErrorException(pg_last_error());
-                $affectedRows = pg_affected_rows($res);
+            $result->setIsSuccessful(true);
+            $result->setMessage('Project deleted successfully.');
```
anarute, line 967: I think you meant `Task deleted successfully`.

model/facade/action/DeleteReportAction.php
```diff
@@ -68,32 +69,42 @@ public function __construct(TaskVO $task) {
...
         $projectVO = DAOFactory::getProjectDAO()->getById($projectId);
-        if (!$projectVO || !$projectVO->getActivation())
-            return -1;
-
-        if ($dao->delete($this->task)!=1) {
-            return -1;
+        if (!$projectVO || !$projectVO->getActivation()) {
+            $result = new OperationResult(false);
+            $result->setErrorNumber(30);
+            $result->setResponseCode(500);
```
anarute, line 102: Maybe 403 here?

anarute (approved):
I've added some minor comments, testing locally it works fine! Thank you

dmtrek14 (approved):
Other than one small nit (which could be addressed later), I think this looks good. Thanks for getting the detailed error messages implemented for tasks - this is a huge improvement.

model/facade/action/DeleteReportAction.php
jaragunde: :+1:

jaragunde (commented):
I've submitted a new patch set amending the existing patches to address your comments. In a new patch I've reviewed all the error codes related to task create, update and delete and make web services pass them instead of using always 500.

web/include/phpreport.css
```diff
@@ -130,6 +130,11 @@ div#tasks textarea {
     margin-bottom: 10px;
 }
 
+/* ExtJS message box */
+.icon-status-error {
```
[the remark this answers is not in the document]
dmtrek14: That's great for the time being! Thanks for updating.

jaragunde:
Thanks for your reviews! I understand it can be merged already.

Issue #110, opened 2016-03-09
Unhandled exception when creating task with missing compulsory data


jaragunde:
[Trac import]
Reported by: jaragunde
Original date: Saturday, 18 September 2010 22:19

I tried to create a task using the service `createTasksService.php`, sending the following (incomplete) data:

```
<tasks sid='r8bsjjj01m9fg1t28hepnfash4'>
    <task><date format='Y-m-d'>2010-8-18</date></task>
</tasks>
```

Instead of answering with an error message, the execution of the service finished because of an internal error. In the error log file there was this message:

```
PHP Fatal error:  Uncaught exception 'SQLQueryErrorException' with message 'ERROR:  23502: null value in column "init" violates not-null constraint
LOCATION:  ExecConstraints, execMain.c:2229' in /var/www/phpreport2/phpreport/model/dao/TaskDAO/PostgreSQLTaskDAO.php:645
Stack trace:
#0 /var/www/phpreport2/phpreport/model/facade/action/CreateReportAction.php(77): PostgreSQLTaskDAO->create(Object(TaskVO))
#1 /var/www/phpreport2/phpreport/model/facade/action/Action.php(112): CreateReportAction->doExecute()
#2 /var/www/phpreport2/phpreport/model/facade/TasksFacade.php(78): Action->execute()
#3 /var/www/phpreport2/phpreport/model/facade/TasksFacade.php(94): TasksFacade::CreateReport(Object(TaskVO))
#4 /var/www/phpreport2/phpreport/web/services/createTasksService.php(216): TasksFacade::CreateReports(Array)
#5 {main}
  thrown in /var/www/phpreport2/phpreport/model/dao/TaskDAO/PostgreSQLTaskDAO.php on line 645
```

jaragunde:
[Trac import]
Comment by: jaragunde
Original date: Saturday, 18 September 2010 23:01

The problem extends to other SQL exceptions, like foreign key constraint violations:
```
PHP Fatal error:  Uncaught exception 'SQLQueryErrorException' with message 'ERROR:  23503: insert or update on table "task" violates foreign key constraint "makeup"
DETAIL:  Key (projectid)=(-1) is not present in table "project".
LOCATION:  ri_ReportViolation, ri_triggers.c:3502' in /var/www/phpreport2/phpreport/model/dao/TaskDAO/PostgreSQLTaskDAO.php:645
Stack trace:
[...]

PR #619, opened 2023-03-06
Error message for missing compulsory fields on task create.

jaragunde:
We check for "not null violation" error messages from PostgreSQL and turn them into a more user friendly message: "Missing compulsory data". We add a specific check and message for missing init and end time fields which are the only ones that realistically can happen at that point; missing project and user fields should have been checked in upper layers.

This fixes an old bug, #110, together with the recent work to turn SQL exceptions into error messages.


anarute (approved):
Looks good to me

dmtrek14:
Looks good to me, as well

PR #621, opened 2023-03-08
Report better errors on user CRUD operations

jaragunde:
This affects the Postgres and LDAP DAOs. Notice there are many operations in the user management screen (assign users to groups and all the user history assignment) that aren't upgraded yet to OperationResult and they keep working as usual.

### DIFF
--- a/model/dao/UserDAO/HybridUserDAO.php
+++ b/model/dao/UserDAO/HybridUserDAO.php
@@ -419,13 +419,16 @@ class HybridUserDAO extends PostgreSQLUserDAO {
      *
      * This function updates the data of a User by its {@link UserVO}.
      * WARNING: this function allows to update to a name that does not exist in LDAP.
-     * Do we really want this behavior?
+     * A user with a name not matching the LDAP won't be able to log in, this should be a
+     * transitory situation while the name is not updated in both places.
      *
      * @param UserVO $userVO the {@link UserVO} with the data we want to update on database.
      * @return OperationResult the result {@link OperationResult} with information about operation status
      */
     public function update(UserVO $userVO) {
-        return parent::update($userVO);
+        $result = parent::update($userVO);
+        $result->setMessage("Warning: user names not matching LDAP won't be able to login.");
+        return $result;
     }
 
     /** User creator for LDAP/PostgreSQL Hybrid.
--- a/model/dao/UserDAO/HybridUserDAO.php
+++ b/model/dao/UserDAO/HybridUserDAO.php
@@ -468,29 +468,10 @@ class HybridUserDAO extends PostgreSQLUserDAO {
      * This function deletes the data of a User by its {@link UserVO}.
      *
      * @param UserVO $userVO the {@link UserVO} with the data we want to delete from database.
-     * @return int the number of rows that have been affected (it should be 1).
-     * @throws {@link SQLQueryErrorException}
+     * @return OperationResult the result {@link OperationResult} with information about operation status
      */
     public function delete(UserVO $userVO) {
-        $affectedRows = 0;
-
-        // Check for a user ID.
-        if($userVO->getId() >= 0) {
-            $currUserVO = $this->getById($userVO->getId());
-        }
-
-        // Otherwise delete a user.
-        if(sizeof($currUserVO) > 0) {
-            $sql = "DELETE FROM usr WHERE id=".$userVO->getId();
-
-            $res = pg_query($this->connect, $sql);
-
-            if ($res == NULL) throw new SQLQueryErrorException(pg_last_error());
-
-            $affectedRows = pg_affected_rows($res);
-        }
-
-        return $affectedRows;
+        return parent::delete($userVO);
     }
 }
 
--- a/model/dao/UserDAO/HybridUserDAO.php
+++ b/model/dao/UserDAO/HybridUserDAO.php
@@ -32,7 +32,7 @@
 include_once(PHPREPORT_ROOT . '/util/SQLIncorrectTypeException.php');
 include_once(PHPREPORT_ROOT . '/util/DBPostgres.php');
 include_once(PHPREPORT_ROOT . '/model/vo/UserVO.php');
-include_once(PHPREPORT_ROOT . '/model/dao/UserDAO/UserDAO.php');
+include_once(PHPREPORT_ROOT . '/model/dao/UserDAO/PostgreSQLUserDAO.php');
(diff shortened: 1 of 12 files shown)


model/dao/UserDAO/HybridUserDAO.php
```diff
@@ -417,35 +418,14 @@ public function getJourneyHistory($userId) {
     /** User updater for LDAP/PostgreSQL Hybrid.
      *
      * This function updates the data of a User by its {@link UserVO}.
+     * WARNING: this function allows to update to a name that does not exist in LDAP.
+     * Do we really want this behavior?
```
jaragunde, line 422: Do we do something about this? This could make sense as a transient state before we change the user in LDAP, and in this state the user would not be able to login. Alternatively, we could check that we rename to another login that exists in LDAP so we would have to modify LDAP first. In any case, this should be a very rarely used operation.

model/dao/UserDAO/HybridUserDAO.php
```diff
@@ -470,29 +468,20 @@ public function create(UserVO $userVO) {
...
-        $res = pg_query($this->connect, $sql);
-
-        if ($res == NULL)
-            if (strpos(pg_last_error(), "unique_usr_login"))
-                throw new SQLUniqueViolationException(pg_last_error());
-            else throw new SQLQueryErrorException(pg_last_error());
+        if($ldapResult["count"] == 0) {
+            $result = new OperationResult(false);
+            $result->setResponseCode(400);
+            $result->setErrorNumber(1234);
```
anarute, line 453: Just double checking if 1234 is correct?

model/dao/UserDAO/HybridUserDAO.php
anarute: I agree that this is such a rare case that we can leave it like this for now.

anarute (approved):
This is great, thank you! Sorry for the late review

model/dao/UserDAO/PostgreSQLUserDAO.php
```diff
@@ -401,38 +401,48 @@ public function getJourneyHistory($userId) {
...
+            $statement = $this->pdo->prepare($sql);
+            $statement->bindValue(":login", $userVO->getLogin(), PDO::PARAM_STR);
+            $statement->bindValue(":id", $userVO->getId(), PDO::PARAM_INT);
+            if (!$emptyPassword)
+                $statement->bindValue(":password", $userVO->getPassword(), PDO::PARAM_STR);
+            $statement->execute();
+
+            $result->setIsSuccessful(true);
+            $result->setMessage('User updated successfully.');
+            $result->setResponseCode(201);
```
dmtrek14, line 427: Nit: If we are updating here and not creating, then we might want to use the 200 response code (for 'okay')

model/dao/UserDAO/PostgreSQLUserDAO.php
```diff
@@ -441,64 +451,104 @@ public function update(UserVO $userVO) {
...
+            $statement = $this->pdo->prepare($sql);
+            $statement->bindValue(":id", $userVO->getId(), PDO::PARAM_INT);
+            $statement->execute();
 
-            if ($res == NULL) throw new SQLQueryErrorException(pg_last_error());
-
-            $affectedRows = pg_affected_rows($res);
+            $result->setIsSuccessful(true);
+            $result->setMessage('User updated successfully.');
+            $result->setResponseCode(201);
```
dmtrek14, line 520: Nit: since we're deleting and not creating, a 200 response code might be better here.

model/dao/UserDAO/PostgreSQLUserDAO.php
```diff
@@ -441,64 +451,104 @@ public function update(UserVO $userVO) {
...
+        }
+        catch (PDOException $ex) {
+            $errorMessage = $ex->getMessage();
+            $resultMessage = "Error updating user:\n";
+            if(strpos($errorMessage, "Foreign key violation")) {
+                $result->setResponseCode(403);
```
dmtrek14, line 526: Nit (again, sorry): My understanding of a 403 is that it means a user doesn't have permission to perform the attempted action. A FK violation feels a little more like a generic 500 to me, but I also recognize that some of these response codes don't always fit every situation. 

Same feedback re the message: you might want to specify that that was an error when deleting the user instead of updating.

dmtrek14 (approved):
I had a few nits, but nothing to hold up the PR over.

model/dao/UserDAO/HybridUserDAO.php
jaragunde: Yes, sorry, I'm not really keeping track of what the error codes are and we aren't really using them. We'd have to follow up with a general review...

model/dao/UserDAO/HybridUserDAO.php
jaragunde: Thanks, I think I'll rewrite the warning message to state the consequences clearly instead of asking this open question.

model/dao/UserDAO/PostgreSQLUserDAO.php
jaragunde: Thanks for catching this, this and the other nits are copy&paste issues I was blind to.

model/dao/UserDAO/PostgreSQLUserDAO.php
jaragunde: You are right about 403 doesn't really fit, but I don't want to use 500 because it's [defined as a "catch-all" error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) for unexpected situations. It's possible to trigger this foreign key violation from the UI, if an admin attempts to delete a user and that user has already created tasks, so I don't consider it an unexpected situation. And I decided not to use 400 because I don't think it's a client error either...

Maybe [409 Conflict](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) is a better match?

jaragunde:
I've pushed fixup patches to address your comments, and a new patch that allows us to report a message from the backend to the UI even in case of success. That's useful to show a warning when updating the login of an LDAP user.

model/dao/UserDAO/PostgreSQLUserDAO.php
anarute: I think 409 makes sense too!

anarute:
The changes look good to me!

jaragunde:
Thanks! I've squashed the fixup patches, and will proceed to merge.

PR #622, opened 2023-03-24
Add support for integration tests and a small set of tests

anarute:
Finally add support for integration tests using docker so we can now test all the service layer without having to change anything in the application design.

The unit tests will work locally without issues, but the integration tests will only work if there is a service running on the port 8000 with the data from the `sql/initialData.sql` file. In that case it's easier to run it with docker, as it has everything setup.

In one terminal build and run the application and the database:

``docker-compose -f docker/docker-compose.test.yml up --build --renew-anon-volumes``

In a second terminal run the ``test`` container to run the tests:

``docker-compose -f docker/docker-compose.test.yml run -T test``

---

I've also added a few tests that check the loginService and task creation.

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -10,21 +10,10 @@ jobs:
   build_and_test:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - name: Validate composer.json and composer.lock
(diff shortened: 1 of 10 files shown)


dmtrek14 (approved):
This is great, @anarute! Thanks for adding these tests. Everything looks good to me.

PR #624, opened 2023-03-31
Add default Zero Hours Day template

anarute:
Make easier for people to skip days by having a default template they can use.

Screenshot from 2023-03-31 17-57-40


dmtrek14 (approved):
Works well for me. Thanks for adding.

PR #625, opened 2023-03-31
Add confirmation modal when deleting template

dmtrek14:
It is very easy to accidentally delete a template. The code here puts a confirmation modal in place so that a user has an escape hatch if they accidentally click the 'delete' button for a template they actually want to keep.


anarute (approved):
Sorry for the delay, it looks good to me

PR #626, opened 2023-04-05
Make sure not to send email to new joiners

anarute:
The script was checking if the user had a journey or not, so if a user joined the current week they would receive a reminder because they are active but haven't filled the past week.

This patch checks if the user joined at least a week ago and for the summary email, also checks if they were active in the previous weeks too before adding them to the lists.


dmtrek14 (approved):
Looks good to me

PR #628, opened 2023-04-17
Add column renderer to handle html in description

dmtrek14:
If someone has added an html element/tag inside their task description, the html gets rendered in the User Tasks report. In order to prevent this, I have added a small renderer function that uses one of Ext.js' native utils - an htmlEncoder. 

To test:
- Add a task with something like the following in the description: "Worked on adding support for <iframe> in XYZ repository"
- Save task
- Under 'Reports', choose 'User tasks'
- Select a date range that encompasses the task you just added
- Check to see if '<iframe>' is simply rendered as the element as text or if there is a blank iframe in the 'Description' column

### DIFF
--- a/web/js/userTasksReport.js
+++ b/web/js/userTasksReport.js
@@ -141,6 +141,11 @@ Ext.onReady(function () {
         return id;
     };
 
+    /* Renderer for html */
+    function descriptionRenderer(text){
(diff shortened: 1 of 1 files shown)


anarute (approved):
That's great, thanks!

jaragunde:
This is great, thank you!

PR #629, opened 2023-04-18
Leave new records empty rather than prefill

dmtrek14:
Records added via the inline edit grids also prefill info and this can be a speedbump since users then have to delete the prefilled text in order to enter their own text. Instead, I propose leveraging the `emptyText` property in the editor which will add placeholder text that a user wouldn't have to delete.

Leaving this as draft for the moment so that folks can try it out to see if they agree with the new approach.

### DIFF
--- a/web/js/areaManagement.js
+++ b/web/js/areaManagement.js
@@ -213,6 +213,7 @@ Ext.onReady(function(){
             editor: {
                 xtype: 'textfield',
                 allowBlank: false,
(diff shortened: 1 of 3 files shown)


anarute (approved):
I have the feeling the user placeholder was added to avoid showing an Error of field required as soon as the Add button is clicked, but I think this solution is still better than having to erase the placeholder in order to add a new user, so it looks good to me.

PR #630, opened 2023-04-19
Display if project has hours logged in the future

anarute:
Sometimes users can mistakenly log hours in the future and this was causing confusion because the total hours vs the hours displayed per user in the report didn't match.

With this patch, if a project has hours logged in the future a field with the amount of hours will be displayed in red in the project details report.

Screenshot from 2023-04-19 13-49-30

### DIFF
--- a/model/vo/ProjectVO.php
+++ b/model/vo/ProjectVO.php
@@ -63,6 +63,7 @@ class ProjectVO {
     protected $movedHours = NULL;
     protected $schedType = NULL;
     protected $type = NULL;
+    protected $futureLoggedHours = NULL;
 
     public function setId($id) {
         if (is_null($id))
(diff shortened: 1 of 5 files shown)


model/vo/ProjectVO.php
dmtrek14, line 1: The only real question I have is whether this code should be in CustomProjectVO.php instead of this file since it is the CustomProjectVO that has the derived values like 'workedHours' and the ProjectVO has just the values that come from the table directly.

dmtrek14 (approved):
See comment attached to the ProjectVO.php. I don't have a strong opinion on where the code goes, honestly - I just noticed that all 'calculated' fields were in another VO that inherits the ProjectVO props. So I won't hold up the PR on where the code is.

model/vo/ProjectVO.php
anarute: That makes sense, I was confused too with what the CustomProjectVO meant vs ProjectVO, I'll change it :)

anarute: Hecho!

PR #632, opened 2023-05-12
Add active property to task type dropdown

anarute:
Allow users to select only the active task types but still show them in the reports or in past tasks.


dmtrek14 (approved):
Looks good to me - thanks!

PR #633, opened 2023-05-30
Account for user journey when calculating days in summary

dmtrek14:
Currently, the Vacation Management page shows Days Booked per Week and includes partial leaves (days where the number of of hours is less than a user journey), but doesn't account for a user booking more vacation hours than their workday hours (journey). This is normally not something to worry about too much since most users will probably use the Vacation Management page to book vacation and that interface takes care of handling the right amount of hours based on journey. However, if the user manually books vacation time via the Tasks page, there could be some days on which the user has more hours booked for vacation than they should. When this happens (more hours booked than should be allowed for a day), the Days Booked per Week does not take the user journey into account and will simply show 1 day booked on a week when it should perhaps show 1.25 days. 

This PR adds accounting for user journey in the calculation of Days Booked per Week and also adds highlighting to days on which a user has booked more vacation hours than their journey.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -161,6 +161,12 @@ define('DAO_BACKEND', 'PostgreSQL');
 define ('YEARLY_HOLIDAY_HOURS', 184);
 
 /**
+ * @name STANDARD_WORKING_DAY
+ * @global int standard working day length (i.e., journey)
+ */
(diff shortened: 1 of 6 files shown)


dmtrek14:
Putting this one on old - there's some unforeseen complexity

PR #634, opened 2023-06-01
Move overlapping task validation to frontend

dmtrek14:
We've been doing validation of overlapping tasks on the backend, but it seems better practice to do this on the frontend where other validation is happening. This code adds the checks for overlapping tasks at the time of validating other values in the frontend js. Additionally, I've made the warnings about overlapping times more informative so that a user knows which tasks overlapped when attempting to save.

### DIFF
--- a/model/dao/TaskDAO/PostgreSQLTaskDAO.php
+++ b/model/dao/TaskDAO/PostgreSQLTaskDAO.php
@@ -739,14 +739,6 @@ class PostgreSQLTaskDAO extends TaskDAO{
      * @return array OperationResult the array of {@link OperationResult} with information about operation status
      */
     public function batchPartialUpdate($tasks) {
-        // if (!$this->checkOverlappingWithDBTasks($tasks)) {
-        //     $result = new OperationResult(false);
-        //     $result->setErrorNumber(10);
-        //     $result->setMessage("Task update failed:\nDetected overlapping times.");
-        //     $result->setResponseCode(400);
-        //     return array($result);
-        // }
-
         $results = array();
         foreach ($tasks as $task) {
             $results[] = $this->partialUpdate($task);
@@ -941,14 +933,6 @@ class PostgreSQLTaskDAO extends TaskDAO{
      * @return array OperationResult the array of {@link OperationResult} with information about operation status
      */
     public function batchCreate($tasks) {
-        // if (!$this->checkOverlappingWithDBTasks($tasks)) {
-        //     $result = new OperationResult(false);
-        //     $result->setErrorNumber(10);
-        //     $result->setMessage("Task creation failed:\nDetected overlapping times.");
-        //     $result->setResponseCode(400);
-        //     return array($result);
-        // }
-
         $results = array();
         foreach ($tasks as $task) {
             $results[] = $this->createInternal($task);
--- a/model/dao/TaskDAO/PostgreSQLTaskDAO.php
+++ b/model/dao/TaskDAO/PostgreSQLTaskDAO.php
@@ -739,13 +739,13 @@ class PostgreSQLTaskDAO extends TaskDAO{
      * @return array OperationResult the array of {@link OperationResult} with information about operation status
      */
     public function batchPartialUpdate($tasks) {
-        if (!$this->checkOverlappingWithDBTasks($tasks)) {
-            $result = new OperationResult(false);
-            $result->setErrorNumber(10);
-            $result->setMessage("Task update failed:\nDetected overlapping times.");
-            $result->setResponseCode(400);
-            return array($result);
-        }
+        // if (!$this->checkOverlappingWithDBTasks($tasks)) {
+        //     $result = new OperationResult(false);
+        //     $result->setErrorNumber(10);
+        //     $result->setMessage("Task update failed:\nDetected overlapping times.");
+        //     $result->setResponseCode(400);
+        //     return array($result);
+        // }
 
         $results = array();
         foreach ($tasks as $task) {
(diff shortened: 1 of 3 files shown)


model/dao/TaskDAO/PostgreSQLTaskDAO.php
```diff
@@ -739,14 +739,6 @@ public function partialUpdate(DirtyTaskVO $taskVO) {
      * @return array OperationResult the array of {@link OperationResult} with information about operation status
      */
     public function batchPartialUpdate($tasks) {
-        if (!$this->checkOverlappingWithDBTasks($tasks)) {
-            $result = new OperationResult(false);
-            $result->setErrorNumber(10);
-            $result->setMessage("Task update failed:\nDetected overlapping times.");
-            $result->setResponseCode(400);
-            return array($result);
-        }
```
anarute, line 748: I would keep the validation in the back end too just to make sure this case is never stored in the database. Does the user get duplicated error messages if we keep it here too?

web/js/tasks.js
```diff
@@ -252,13 +252,45 @@ function removeFreshEmptyTask() {
...
+            if(getMinutes(t.data.endTime) > getMinutes(x.data.initTime) && getMinutes(t.data.initTime) < getMinutes(x.data.endTime)){
+                overlapsTasks = true;
+                overlapping.push(x);
+                message = "Task from " + x.data.initTime + " to " + x.data.endTime + " overlaps with task from " + t.data.initTime + " to " + t.data.endTime;
+            }
+            else if (getMinutes(t.data.initTime) == getMinutes(x.data.initTime) || getMinutes(t.data.endTime) == getMinutes(x.data.endTime)){
+                overlapsTasks = true;
+                overlapping.push(x);
+                message = "Task from " + x.data.initTime + " to " + x.data.endTime + " overlaps with task from " + t.data.initTime + " to " + t.data.endTime;
+            }
```
anarute, line 280: I wonder if we can merge these blocks since they yield the same thing?

```suggestion
            if((getMinutes(t.data.endTime) > getMinutes(x.data.initTime) && getMinutes(t.data.initTime) < getMinutes(x.data.endTime)) || (getMinutes(t.data.initTime) == getMinutes(x.data.initTime) || getMinutes(t.data.endTime) == getMinutes(x.data.endTime))){
                overlapsTasks = true;
                overlapping.push(x);
                message = "Task from " + x.data.initTime + " to " + x.data.endTime + " overlaps with task from " + t.data.initTime + " to " + t.data.endTime;
            }
```

Maybe we can improve readability by storing the checks in variables:

```suggestion
            let overlapsTasksDuration = (getMinutes(t.data.endTime) > getMinutes(x.data.initTime) && getMinutes(t.data.initTime) < getMinutes(x.data.endTime));
            let coincidesInitOrEndTimes =  (getMinutes(t.data.initTime) == getMinutes(x.data.initTime) || getMinutes(t.data.endTime) == getMinutes(x.data.endTime));
            if (overlapsTasksDuration || coincidesInitOrEndTimes) {
                overlapsTasks = true;
                overlapping.push(x);
                message = "Task from " + x.data.initTime + " to " + x.data.endTime + " overlaps with task from " + t.data.initTime + " to " + t.data.endTime;
            }
```

anarute: I noticed the message will be overwritten and show only the latest overlapping tasks, is it on purpose? It got me thinking if we want to show them all or only one, I don't have a strong opinion.

model/dao/TaskDAO/PostgreSQLTaskDAO.php
dmtrek14: Yeah, this could work. The frontend will validate first and if not valid will not hit the backend. So, if for some reason the frontend validation is borked, then the backend would send the same message it always had. We'd want to test and see what it actually does. I will work on that.

web/js/tasks.js
```diff
@@ -995,7 +1045,7 @@ Ext.onReady(function(){
     // Implement autosave of data, when valid contents are typed in to the task fields
     window.setInterval(function () {
         if(isUnsaved()) {
-            saveTasks(false);
+            saveTasks(true);
```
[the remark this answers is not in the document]
dmtrek14: Ooo, maybe.

web/js/tasks.js
dmtrek14: Oops. Not on purpose. But I don't have a strong opinion since if there is more than one problem, a user will have to fix the first failure and simply get another failure for the next problem until they fix everything. I can try to get it to show all, since that was my original intent.

dmtrek14: The suggestions on readability are good. I like the idea of storing in variable to make it easier to parse.

model/dao/TaskDAO/PostgreSQLTaskDAO.php
dmtrek14: I tested quite a bit and the front is catching everything I throw at it so far, so I haven't seen what it would look like if a task got through. But it's still a good idea to restore the backend check. I've done so and pushed.

tests/integration/TaskServiceTest.php
```diff
@@ -27,17 +27,4 @@ public function testCreateTask(): void
...
-        // Creating a task with the same values of the previous test should fail
-        // because they overlap.
-        $request = '<?xml version="1.0" encoding="ISO-8859-15"?>';
-        $request .= '<tasks sid="'. $this->sessionId .'">';
-        $request .= '<task><date>2023-03-24</date><initTime>00:00</initTime><endTime>03:00</endTime>';
-        $request .= '<customerId>1</customerId><projectId>1</projectId></task></tasks>';
-        $res = $this->makeRequest('web/services/createTasksService.php', null, 'POST', $request);
-        $this->assertStringContainsString('Task creation failed', (string)$res->errors->error);
-        $this->assertStringContainsString('Detected overlapping times.', (string)$res->errors->error);
-    }
```
anarute, line 42: Since we are not removing the check from the backend anymore, I think we should keep the test

dmtrek14: Re-added 😄

anarute (approved):
Looks good to me, thanks!

PR #635, opened 2023-06-19
Remove code related to deprecated Request entity

anarute:
The Request VO and related methods were deprecated a long time ago, this code is probably a leftover from it. It was not being used anywhere else in the project.

Commit that removed the Request entity a28cf41b


dmtrek14 (approved):
Looks good to me.

PR #637, opened 2023-06-22
Convert configuration to use .env

anarute:
Stop using `config.php`  and start using `.env` so it can be used by different stacks in the project and Docker.

Also get rid of the `test` container and add the clean tasks method before tests.


dmtrek14 (approved):
Looks good!!

PR #638, opened 2023-06-26
Add Alembic and FastAPI to the project

anarute:
We decided to use Alembic to manage our database migrations from now on and FastAPI to build our backend API endpoints.

This PR adds the basic structure for that:

- Port all the database models
- Create initial DB migrations
- Add FastAPI architecture and projects list endpoint to be used as example


dmtrek14 (approved):
Was able to run migrations as expected and API run correctly and brings back results to test page.

PR #640, opened 2023-07-03
Add week number variable to email subject

anarute:
These values were hard coded in the config file.


.env.example
```diff
@@ -45,10 +45,10 @@ NO_FILL_TEMPLATE_WARNING="templates/no_fill_warning.txt"
 NO_FILL_TEMPLATE_CRITICAL="templates/no_fill_critical.txt"
 NO_FILL_TEMPLATE_LAST="templates/no_fill_last.txt"
 NO_FILL_TEMPLATE_MANAGERS="templates/no_fill_summary_managers.txt"
-NO_FILL_SUBJECT_WARNING="Please log your hours in PHPReport"
-NO_FILL_SUBJECT_CRITICAL="[URGENT] Log your hours in PHPReport"
-NO_FILL_SUBJECT_LAST="[URGENT] Three days remain to update PHPReport"
-NO_FILL_SUBJECT_MANAGERS="Report: Summary of people late filling in PHPReport"
+NO_FILL_SUBJECT_WARNING="Please log your hours in PhpReport to complete ###WEEK_NUMBER###"
+NO_FILL_SUBJECT_CRITICAL="[URGENT] Log your hours in PhpReport to complete ###WEEK_NUMBER###"
```
dmtrek14, line 49: Would the subject make more sense if the word 'week' was added in front of the week number for each of these variables?

dmtrek14 (approved):
One small nit for clarity - take it or leave it. Otherwise looks good.

.env.example
anarute: oh good catch! It should be W+the number instead, I forgot to add it, thank you :)

anarute: Done

PR #641, opened 2023-07-04
Improve how to get env file and deprecate custom DAOs

anarute:
- The env file was being loaded on every request, making the application slower.

- The custom DAO logic wasn't being used and only added complexity and unnecessary
checks on every request.


dmtrek14 (approved):
Looks good.

Issue #494, opened 2021-03-16
PhpReport login page should redirect if already logged in


jaragunde:
The login page is setup to redirect you to tasks.php (or any page that was set as the `login_success_url` GET variable) after a successful login.

We should also perform the same relocation if a user enters this page (e.g. by typing its URL) when it's already logged in. The index page redirects to login.php, which makes this issue more visible.

AtmegaBuzz:
Hey @jaragunde, I'm new to this org and want to contribute to it. Can you assign this to me

anarute:
@AtmegaBuzz since this issue is a bit old, I'm trying to reproduce it and it doesn't seem to be an issue anymore. @jaragunde is on leave so he may take a while to reply, I would wait for him to avoid spending time on something that is not a real issue anymore, but thanks for volunteering!

AtmegaBuzz:
@anarute I ran the main branch in my machine and it seem to be happening on my end. after logging in if i go back to base url localhost:8000 it redirects me to login.php not to tasks.php

AtmegaBuzz:
Sure I would like to work on it you can assign it to me, Thanks @dmtrek14

anarute:
@AtmegaBuzz You're right this behavior was happening indeed and I couldn't see it because we use a different authentication method. Thank you for the fix!

PR #644, opened 2023-07-05
fix login page redirect if already logged in

AtmegaBuzz:
Solves: PhpReport login page should redirect if already logged in #494

What i did:
Added a logic check in login.php to check if LoginManager.isLogged is not returning null. 
If null is not returned then the the user is redirected to tasks.php

### DIFF
--- a/web/login.php
+++ b/web/login.php
@@ -37,10 +37,6 @@ function set_location_header_on_login_success_url(){
 /* First check if a custom authentication header was set. */
 $customHeader = ConfigurationParametersManager::getParameter('EXTERNAL_AUTHENTICATION_USER_HEADER');
 
-if (LoginManager::isLogged()) {
-    set_location_header_on_login_success_url();
-}
-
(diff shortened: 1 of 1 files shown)


web/login.php
```diff
@@ -36,6 +36,11 @@ function set_location_header_on_login_success_url(){
 
 /* First check if a custom authentication header was set. */
 $customHeader = ConfigurationParametersManager::getParameter('EXTERNAL_AUTHENTICATION_USER_HEADER');
+
+if (LoginManager::isLogged()) {
```
dmtrek14, line 40: I have a small nitpick that is mostly stylistic. Since we already have a `if/else if` block of checks regarding the login status, perhaps your check would make more sense as an `else if` block at the end of the current block.

AtmegaBuzz: Done

dmtrek14 (approved):
Thanks for making the change. I'll merge it in.

PR #647, opened 2023-07-10
Add JWT auth support

anarute:
Add support to receive a Bearer token in the API requests and decode it.


dmtrek14 (approved):
Looks good to me.

PR #648, opened 2023-07-13
Add Task Type entity

anarute:
This will be used in the task form and in the future we'll have a UI so admins can manage this better.

### DIFF
--- a/api/models/timelog.py
+++ b/api/models/timelog.py
@@ -24,7 +24,7 @@ class Task(Base):
     story = Column(String(length=80), nullable=True)
     telework = Column(Boolean, nullable=True)
     text = Column(String(length=8192), nullable=True)
-    ttype = Column(String(length=40), nullable=True)
+    task_type = Column("ttype", String, ForeignKey("task_type.slug"), nullable=True)
     phase = Column(String(length=15), nullable=True)
     onsite = Column(Boolean, default=False, nullable=False)
     updated_at = Column(postgresql.TIMESTAMP(), nullable=True)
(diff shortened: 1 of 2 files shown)


api/models/timelog.py
```diff
@@ -24,7 +24,7 @@ class Task(Base):
     story = Column(String(length=80), nullable=True)
     telework = Column(Boolean, nullable=True)
     text = Column(String(length=8192), nullable=True)
-    ttype = Column(String(length=40), nullable=True)
+    task_type = Column("ttype", String, ForeignKey("task_type.slug"), nullable=True)
```
dmtrek14, line 27: Probably silly question: why use task_type.slug as the foreign key instead of the task_type.id?

anarute: This is to be a bit less disruptive with the current state. Since in the current column it already saves the slug the values would be pretty much the same.

I also thought that for queries that don't require the full name we wouldn't need join with the task type table in order to get the type of the task.

anarute: If it makes more sense we could make the slug the primary key and remove the id column?

dmtrek14: Ah, of course! This makes sense. It'll work fine as-is; I think we could leave the id column in case we ever wanted to go another direction.

anarute:
@dmtrek14 before merging I noticed some inconsistencies between the model and the final migration that should be fixed now, but would you mind doing a second look please? just to make sure I didn't forget anything.

dmtrek14:
@anarute Look good here.

PR #649, opened 2023-07-19
Add React frontend with authentication

dmtrek14:
- Initial add of React frontend
- Add audience to Python jwt handler
- Implement OIDC auth in React frontend 
- Handle OIDC auth in the php side  

These changes add a React frontend that will be the start of new pages in the application. OIDC authentication has been added as a context in the React app so that the app will have the user/token information everywhere - this makes it fairly straightforward to hit the new Python API with a bearer token. The existing php login manager was also changed in order to handle OIDC authentication. 

Example pages have been set up - both in the React app and the php app - showing how to use the token from the OIDC in an Authorization header as a bearer token. On the php side, it is `fastapiTest.php` and on the React side, it is `Dedications.jsx`.


.env.example
```diff
@@ -23,6 +23,7 @@ USE_EXTERNAL_AUTHENTICATION=false
 EXTERNAL_AUTHENTICATION_USER_HEADER=""
 JWT_ALGORITHM=
 JWT_SECRET=
+JWT_AUDIENCE=
```
anarute, line 26: Can we call it JWT_CLIENT_ID instead to make it clearer what it is?

anarute:
Since this wasn't merged yet, I would take the opportunity to squash the refactor commit with the first commit that you introduce the React app. Maybe changing the order will make it a bit easier to rebase:
-  [Add audience to python jwt handler](https://github.com/Igalia/phpreport/pull/649/commits/b8f2a34a921c26de541c0de9ce56f5f940ad3e37)
- Commit that adds the React app
- [Handle OIDC in php side](https://github.com/Igalia/phpreport/pull/649/commits/78703494727adfa4d961c2fe826d91f05f63ced3)

Also adding a better message in the squash commit that describes better what it does, like the description of this pull request.

config/permissions.php
```diff
@@ -142,12 +188,14 @@
...
-        '/projectDetailsReport.php', '/projectDetails.php',
+        '/projectDetailsReport.php',
+        '/projectDetails.php',
         '/services/getProjectUserCustomerReportJsonService.php',
         //user details report: check any user
         '/viewUserDetails.php',
         '/services/getUserProjectCustomerReportJsonService.php'
     )
-);
+);
```
anarute, line 201: Can we have a commit just for the code styling changes? To avoid mixing it with the real changes.

src/index.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8" />
+  <!-- <link rel="icon" type="image/svg+xml" href="/vite.svg" /> -->
```
anarute, line 6:
```suggestion
```

src/.env.example
```diff
@@ -0,0 +1,7 @@
+VITE_OIDC_AUTHORITY=
+VITE_OIDC_CLIENT_ID=
+VITE_OIDC_CLIENT_SECRET=
+VITE_OIDC_REDIRECT_URL=
+VITE_OIDC_METADATA_URL=
+VITE_OIDC_RESPONSE_TYPE=
+VITE_API_BASE='http://localhost:8555'
```
anarute, line 7: Why do the variable names have VITE in front of it?

.env.example
dmtrek14: We could. I had named it audience since the Python piece was looking for audience and it was clear there. I'm open either way.

config/permissions.php
dmtrek14: I'm backing out any formatting changes for the moment, as they are distracting and not really necessary

anarute (approved):

anarute:
Looks good to me, thanks!

PR #652, opened 2023-07-25
Change property that gets user from OIDC provider

anarute:
When switching to keycloak, the property used to fetch the username is different.


dmtrek14 (approved):
LGTM, thanks!

PR #655, opened 2023-07-31
Move images, svgs, favicons to assets folder

dmtrek14:
Eventually all assets will live within `frontend`, but this cleans up the projects a little until that time.


anarute (approved):
Looks good to me

PR #657, opened 2023-08-02
Update auth handler, login manager, and env

dmtrek14:
- Add more props to env: since OIDC providers may vary on what the property name is for username, it seems wise to make that a variable; add certs endpoint for token validation; separate client-id and audience, as those might vary, as well
- Update auth handler to use certs endpoint to validate token
- Clean up login manager


api/auth/auth_handler.py
```diff
@@ -1,17 +1,21 @@
...
 
 def token_response(token: str):
     return {"access_token": token}
 
 
 def decode_token(token: str) -> dict:
-    decoded_token = jwt.decode(token, JWT_SECRET, audience=JWT_AUDIENCE, algorithms=[JWT_ALGORITHM])
+    jwks_client = PyJWKClient(OIDC_CERTS_URL)
+    signing_key = jwks_client.get_signing_key_from_jwt(token)
+    decoded_token = jwt.decode(token, signing_key.key, audience=JWT_AUDIENCE, algorithms=[JWT_ALGORITHM])
```
anarute, line 20: Does it make sense to replace JWT_AUDIENCE for OIDC_CLIENT_ID here as well since they are the same value? And then get rid of JWT_AUDIENCE

anarute (approved):
Other than the very minor suggestion, it looks good to me, thanks!

api/auth/auth_handler.py
dmtrek14: They aren't always the same, apparently. In the decoded access_token from Keycloak, for example, the `aud` is actually 'account' and then the client-id is elsewhere in the json object. So, I think we may want to leave it for now.

anarute: Oh I see, got it, thanks!

PR #664, opened 2023-09-11
Remove vendor folders from dockerignore

anarute:
They are needed in order to run the apps - it was working in some cases where the dev had run composer install or npm install manually.


dmtrek14 (approved):
Looks good

PR #665, opened 2023-09-11
Increase login column size to account for longer usernames

dmtrek14:
The 'login' column was only set to 20 characters, which seemed too small, so the size has been increased to 100 characters.


anarute (approved):
Looks good to me!

PR #667, opened 2023-09-12
Add template endpoint

dmtrek14:
First, there are some small changes to the `template` table to allow for storage of global templates in the database.

Second, this adds a GET endpoint to the API to retriever user templates. This will get all templates the user has created as well as all global templates.

### DIFF
--- a/api/models/timelog.py
+++ b/api/models/timelog.py
@@ -63,13 +63,14 @@ class Template(Base):
     story = Column(String(length=80), nullable=True)
     telework = Column(Boolean, nullable=True)
     onsite = Column(Boolean, nullable=True)
-    text = Column(String(length=8192), nullable=True)
-    ttype = Column(String(length=40), nullable=True)
+    description = Column("text", String(length=8192), nullable=True)
+    task_type = Column("ttype", String(length=40), nullable=True)
     init_time = Column(Integer, nullable=True)
     end_time = Column(Integer, nullable=True)
-    customer = Column("customerid", Integer, ForeignKey("customer.id"), nullable=True)
-    user = Column("usrid", Integer, ForeignKey("usr.id"), nullable=False)
-    project = Column("projectid", Integer, ForeignKey("project.id"), nullable=True)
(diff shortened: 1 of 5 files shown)


anarute (commented):
Apart from the minor suggestion, it looks good to me!

Also another nit is that in the first commit it is written that some columns were deprecated, but I don't see that in the code.

api/models/timelog.py
```diff
@@ -63,13 +63,14 @@ class Template(Base):
...
+    task_type = Column("ttype", String(length=40), nullable=True)
     init_time = Column(Integer, nullable=True)
     end_time = Column(Integer, nullable=True)
-    customer = Column("customerid", Integer, ForeignKey("customer.id"), nullable=True)
-    user = Column("usrid", Integer, ForeignKey("usr.id"), nullable=False)
-    project = Column("projectid", Integer, ForeignKey("project.id"), nullable=True)
+    customer_id = Column("customerid", Integer, ForeignKey("customer.id"), nullable=True)
+    user_id = Column("usrid", Integer, ForeignKey("usr.id"), nullable=True)
+    project_id = Column("projectid", Integer, ForeignKey("project.id"), nullable=True)
+    is_global = Column(Boolean, nullable=False)
```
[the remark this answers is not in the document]
dmtrek14: I had started by removing the telework and onsite columns, but that was going to break some things

api/models/timelog.py
```diff
@@ -70,7 +70,7 @@ class Template(Base):
     customer_id = Column("customerid", Integer, ForeignKey("customer.id"), nullable=True)
     user_id = Column("usrid", Integer, ForeignKey("usr.id"), nullable=True)
     project_id = Column("projectid", Integer, ForeignKey("project.id"), nullable=True)
-    is_global = Column(Boolean, nullable=False)
+    is_global = Column(Boolean, nullable=False, default=False)
```
anarute, line 73: Sorry for the nit, but I think this belongs to the previous commit.

dmtrek14: Never be sorry for nits ;) Should be fixed now

PR #668, opened 2023-09-13
Improve initial setup

anarute:
Several improvements on how we can setup the development environment using docker and a clean database.

### DIFF
--- a/docs/src/developer/devel-setup.md
+++ b/docs/src/developer/devel-setup.md
@@ -26,13 +26,23 @@ folder of the project with:
 
 `docker-compose -f docker/docker-compose.dev.yml up`
 
-To run migrations:
+During the first time running the app, the API will fail because it needs to
+access the database tables that still don't exist, so run the migrations
+to create the tables with:
 
-`docker exec -ti phpreport-db alembic upgrade head`
+`docker exec -ti phpreport-api alembic upgrade head`
 
-To create migrations:
+Kill the containers and run them again, so the API can connect to the database
+correctly.
 
-`docker exec -ti phpreport-db alembic revision --autogenerate -m "Migrations description"`
+[Optional] When running the project for the first time, you can add some
+basic data for the development environment:
+
+`docker exec -dti phpreport-db psql phpreport -U phpreport < sql/initialData.sql`
+
+If you make changes to the models and want to create migrations:
+
+`docker exec -ti phpreport-api alembic revision --autogenerate -m "Migrations description"`
 
 All the services are setup to reload when the files are updated without the need
 of rebuild the containers.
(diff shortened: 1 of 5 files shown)


docs/src/developer/devel-setup.md
```diff
@@ -26,13 +26,23 @@ folder of the project with:
...
 
-`docker exec -ti phpreport-db alembic upgrade head`
+`docker exec -ti phpreport-api alembic upgrade head`
 
-To create migrations:
+Kill the containers and run them again, so the API can connect to the database
+correctly.
 
-`docker exec -ti phpreport-db alembic revision --autogenerate -m "Migrations description"`
+[Optional] When running the project for the first time, you can add some
```
dmtrek14, line 38: I'm not sure about marking this as optional. It feels to me that some of the data is a little less optional. The user groups are probably something we'd want to have folks add. And at least one user record is probably also recommended. The data from about line 52 on in the initialData.sql file seem like the most optional of the bunch.

In the migration for adding the task types table, you nicely added an insert statement for some default task types. I wonder if it would be worth updating the initial migration to add the bare metal data needed. Then modify the initialData.sql to include only the optional data (and perhaps rename to optionalInitialData.sql). What do you think?

dmtrek14 (commented):
Good changes here to simplify setup - thanks! I just had the one question related to what is/isn't optional in terms of initial data. 

Another thought: now that we're using alembic for migrations and you have modified the docker-compose file, can we actually get rid of the `test.db.Dockerfile`? If yes, then we could probably also get rid of a few of the sql files we were copying into the container.

docs/src/developer/devel-setup.md
anarute: > can we actually get rid of the test.db.Dockerfile?

Not yet, it's still being used for the automated tests: https://github.com/Igalia/phpreport/blob/main/.github/workflows/runtests.yml#L17

> I'm not sure about marking this as optional. It feels to me that some of the data is a little less optional. The user groups are probably something we'd want to have folks add. And at least one user record is probably also recommended. The data from about line 52 on in the initialData.sql file seem like the most optional of the bunch.

> In the migration for adding the task types table, you nicely added an insert statement for some default task types. I wonder if it would be worth updating the initial migration to add the bare metal data needed. Then modify the initialData.sql to include only the optional data (and perhaps rename to optionalInitialData.sql). What do you think?

I'm not sure about adding them to the migrations because even though they are needed for the app to work they should be configured per environment, I'm thinking on the case of setting up a production env for instance. Maybe what we want as a next improvement is a script to insert seed data for dev and production environments? I think only the user groups are more mandatory for any environment. I'm a bit concerned to add hard coded users for any environment. In the case of the task types I've added there because they were hardcoded in the front end, but maybe when we have a seed script it makes sense to also remove them from there?

dmtrek14: I like the idea of seed data based on environments as a next improvement and moving the Task Type data into one of those.

dmtrek14 (approved):

PR #669, opened 2023-09-13
Replace Vite project with NextJs

negreirosleo:
This PR replaces the current implementation on Vite for a NextJs implementation.

It doesn't have a lot besides some boilerplate code and the auth.


docker/docker-compose.dev.yml
dmtrek14, line 1: Nit: looks like there are some spacing discrepancies here in the `volumes` section that could be cleaned up.

dmtrek14 (commented):
This is a pretty straight-forward drop-in replacement of vite, so there wasn't much that stood out in terms of changes. Everything I have here is nitpicks.

docker/docker-compose.dev.yml
anarute: Actually I think it was wrong before and now it's on the same level as the others.

anarute:
Agree with Danielle on the suggestions, other than that it looks good to me :)

docker/docker-compose.dev.yml
negreirosleo: they had only 1 space of identation and I changed it to 2

PR #670, opened 2023-09-15
Add pagination and filter to the projects endpoint

anarute:
- Implement projects basic pagination with offset and limit query params
- Add project status filter that allows the user to filter for active and inactive projects, if there's no filter all the projects will be returned.

### DIFF
--- a/api/services/projects.py
+++ b/api/services/projects.py
@@ -5,5 +5,10 @@ from models.project import Project
 
 
 class ProjectService(AppService):
-    def get_items(self) -> List[Project]:
-        return self.db.query(Project).all() or []
+    def get_items(self, offset, limit, status) -> List[Project]:
+        query = self.db.query(Project)
+        if status == "active":
+            query = query.filter(Project.activation is True)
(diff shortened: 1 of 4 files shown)


api/services/projects.py
```diff
@@ -5,5 +5,10 @@
 
 
 class ProjectService(AppService):
-    def get_items(self) -> List[Project]:
-        return self.db.query(Project).all() or []
+    def get_items(self, offset, limit, status) -> List[Project]:
+        query = self.db.query(Project)
+        if status == "active":
+            query = query.filter(Project.activation == True)
```
dmtrek14, line 11: I ran into a build failure with checking bools, too, for the templates endpoint. For flake8 to be happy, it expects the following, I think.
`query = query.filter(Project.activation is True)`
Same for line 13.

anarute: Solved, thank you. I wonder why locally is passing, I'll check later, probably different versions.

dmtrek14 (approved):

PR #674, opened 2023-09-20
Create ContentSidebar

negreirosleo:
This PR creates the right sidebar and adds responsive styles for both of the sidebars.

The responsive part of the ContentSidebar is a bit funky right now but it will be adjusted when we have more content on the main page and on the sidebar.

[Screencast from 2023-09-20 18-12-48.webm](https://github.com/Igalia/phpreport/assets/20492786/7aa2fdfb-b9c6-458e-9bd1-f03d89526d08)

### DIFF
--- a/frontend/src/app/globals.css
+++ b/frontend/src/app/globals.css
@@ -94,6 +94,10 @@ body {
       rgb(var(--background-end-rgb))
     )
     rgb(var(--background-start-rgb));
+
+  @media (max-width: 600px) {
+    flex-direction: column;
+  }
 }
 
 a {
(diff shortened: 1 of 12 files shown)


dmtrek14 (commented):
Now that you have added the right panel, it might be a good time to reposition the 'Skip navigation' link. It now gets cut off by the right panel. 
Screenshot from 2023-09-20 16-01-21

What about centering the link? It feels like that might be the safest place for it. As long as we leave a little white space at the top of the main section before the form and other components, it should show up okay and not crowd the other content too much. 

Other than that small nit and knowing that you'll continue to refine as we add more content, this is looking nice!

dmtrek14 (approved):
Looks good - thanks for the adjustment

anarute (approved):
Looks good to me!

PR #678, opened 2023-09-21
Add POST endpoint for templates

### LINKED ISSUE #676: Add endpoint for creating templates
Add POST endpoint for templates.

Templates have only one mandatory field which is the name.

If a template is global they shouldn't have a user id.

dmtrek14:
Closes #676

The changes here add the necessary schema, service methods, and endpoint for the creation of templates.

- Db connection updated so expire on commit is false. This prevents errors when trying to refresh the object for return to the api from service
- User model imported in timelog model explicitly. This solves occasional 'NoneType' error being thrown. I think it is due to the fact that we have foreign keys to user, which has a model defined in a separate file.
- Add template schema. Includes validation for column size, as well as check for is_global and user_id.
- Add template endpoint. Also add a 404 as a potential response type for the router.
- Add service to create template

Notes: For the schemas, I used the full-stack-fastapi-postgresql as a loose guide. Shared properties are all on the base schema as optional. Then schemas that need other properties or for properties to be mandatory inherit and override the base.

### DIFF
--- a/api/models/timelog.py
+++ b/api/models/timelog.py
@@ -13,6 +13,7 @@ from sqlalchemy.dialects import postgresql
 from sqlalchemy.orm import relationship, Mapped
 from sqlalchemy.ext.hybrid import hybrid_property
 from models.project import Project
+from models.user import User
 
 from db.base_class import Base
 
@@ -79,7 +80,7 @@ class Template(Base):
     init_time = Column(Integer, nullable=True)
     end_time = Column(Integer, nullable=True)
     customer_id = Column("customerid", Integer, ForeignKey("customer.id"), nullable=True)
-    user_id = Column("usrid", Integer, ForeignKey("usr.id"), nullable=True)
+    user_id = Column("usrid", Integer, ForeignKey(User.id), nullable=True)
     project_id = Column("projectid", Integer, ForeignKey("project.id"), nullable=True)
     is_global = Column(Boolean, nullable=False, default=False)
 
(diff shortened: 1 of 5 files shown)


api/schemas/timelog.py
```diff
@@ -12,23 +12,46 @@ class Config:
...
     user_id: Optional[int]
     project_id: Optional[int]
-    is_global: bool
+    is_global: Optional[bool]
+
+    @root_validator(pre=True)
+    def user_template_not_global(cls, values):
+        user_id, is_global = values.get("user_id"), values.get("is_global")
+        if is_global is not None and user_id is not None and is_global is True:
+            raise ValueError("Global templates should not have a user_id.")
```
anarute, line 32: The check can be simplified a bit and we also need to cover when is not global and doesn't have user, to avoid having "abandoned" templates:

```suggestion
        if is_global and user_id is not None:
            raise ValueError("Global templates should not have a user_id.")
        elif not is_global and not user_id:
            raise ValueError("Private templates should have a user_id.")
```

dmtrek14: Great point. Fixing.

api/models/timelog.py
```diff
@@ -13,6 +13,7 @@
 from sqlalchemy.orm import relationship, Mapped
 from sqlalchemy.ext.hybrid import hybrid_property
 from models.project import Project
+from models.user import User  # noqa: F401
```
anarute, line 16: We can avoid excluding this rule by using the import in the model:

`    user_id = Column("usrid", Integer, ForeignKey(User.id), nullable=True)`

anarute (approved):
It looks good to me, thank you for the changes!

PR #680, opened 2023-09-26
Create task form

negreirosleo:
### Changes

This PR creates the  new Task form, it:
- adds react-query to fetch and manage the data
- creates the /task page
- creates a `useForm` hook to manage the form state
- creates the components: Input, Select, TextArea with floating label

In the Adobe Layout the form items have 40px of height, but I changed it to 56px to accommodate the floating label.

### Screenshots

image


anarute:
I'm tyring to test it locally and I keep being redirected to /web/v2 when I try to access /web/v2/tasks
Is there something I need to do to test it properly?

dmtrek14:
@anarute I also keep getting redirected, but I tried commenting out the `signinRedirect` in the `AuthorizedPage.tsx` and that let me get to the form.

A few small things:
It makes sense to make the inputs a little larger to accommodate the labels, but it does make the buttons feel a little off. Do you plan on increasing the button size to match the height? Or vertically center the buttons to align with the 'More actions' dropdown? I think if you increase the button height, you might also want to use `--joy-fontSize-md` for the font size in the buttons instead of the smaller text to make the button feel more proportioned.

The clear button does not seem to clear the task description.

On mobile, the 'More actions' and the buttons are a bit squished on one line, so perhaps the buttons should flow underneath the dropdown at small screen sizes.

anarute:
I agree with Danielle's comment and would also add that I think we should make the dropdowns have autocomplete feature from the beginning, as keyboard navigation was the number one request from our users.

anarute:
It seems the autocomplete clean up is not working well, once I select a value I cannot remove it:

[Screencast from 2023-09-28 13-51-31.webm](https://github.com/Igalia/phpreport/assets/333447/8c731344-a6e9-44a0-b684-dbc1592bb091)

frontend/src/ui/Select/Select.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+type Option = {
+  value: string
+  label: string
+}
+
+type SelectProps = {
+  label: string
+  value?: string
+} & Omit<AutocompleteProps<Option, undefined, undefined, undefined>, 'value'>
```
anarute, line 15: What do these undefined mean here?

negreirosleo: They come from the `AutocompleteProps`:

```ts
interface AutocompleteComponent {
    <T, Multiple extends boolean | undefined = undefined, DisableClearable extends boolean | undefined = undefined, FreeSolo extends boolean | undefined = undefined>(props: AutocompleteProps<T, Multiple, DisableClearable, FreeSolo>): JSX.Element;
    propTypes?: any;
}
```

We have the generic `T` (which is the `Option` type in our case), and then we have: Multiple, DisableClearable and FreeSolo, but we aren't using any of those props in the component, so they are all undefined.

negreirosleo: here is the signature in the .d.ts file:
```ts
/**
 * AutocompleteProps signature:
 * @template T
 * @param {string | object} T The option structure, must be a string or an object (by default, only accepts objects with { label: string } )
 * @param {boolean | undefined} Multiple If your component is set with property multiple as true
 * @param {boolean | undefined} DisableClearable If your component is set with property disableClearable as true
 * @param {boolean | undefined} FreeSolo If your component is set with property freeSolo as true
 */
 ```

Comment (negreirosleo) [reply frontend/src/ui/Select/Select.tsx:15]: I think it could also be: `AutocompleteProps<Option, boolean, boolean, boolean>` but I don't think it would change anything, I'm just taking those props to extend it to the `Select` component so we can use the same API

Comment (anarute) [reply frontend/src/ui/Select/Select.tsx:15]: Got it, thank you!

**Comment by anarute** [comment]
I think for this PR, adding the To and From fields should be enough, and then the timer, template and data fetching can be in new PRs, what do you think @dmtrek14 ?

**Comment by dmtrek14** [comment]
> I think for this PR, adding the To and From fields should be enough, and then the timer, template and data fetching can be in new PRs, what do you think @dmtrek14 ?

I agree. It will help separate the work out a little better. We expect the timer to be a little complex and it makes sense for it to be its own PR.

File: frontend/src/app/tasks/hooks/useTaskTypes.ts [hunk 2]
```diff
@@ -0,0 +1,36 @@
+import { useQuery } from '@tanstack/react-query'
+import { useAuth } from 'react-oidc-context'
+
+type Task = {
```
dmtrek14, line 4: This should probably be named `TaskType` to avoid confusion with the tasks later on.

frontend/src/app/tasks/hooks/useTaskTypes.ts
```diff
@@ -0,0 +1,36 @@
...
+  slug: string
+  name: string
+  active: boolean
+}
+
+// Temporary disable so we don't need to change the fetch api for now.
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const fetchTasks = (token: string): Promise<Array<Task>> => {
```
dmtrek14, line 12: Same nit here: it might be clearer to name this `fetchTaskTypes` to avoid confusion between a task type and a task later on

frontend/src/app/tasks/hooks/useTaskTypes.ts
```diff
@@ -0,0 +1,36 @@
...
+  ])
+  return mockData
+}
+
+export const useTaskTypes = () => {
+  const auth = useAuth()
+  const token = auth.user?.access_token || ''
+
+  const { data } = useQuery({
+    queryKey: ['tasks', token],
```
dmtrek14, line 28: Same thing: might be best to use `taskTypes`for the query key since we will have `tasks` soon

frontend/src/app/tasks/page.tsx
dmtrek14, line 1: The designer updated the prototype a little recently, which included a slightly different order for the fields on the form. The new order is like so:
Project
Start timer | From | To
Logged time 
Task description
Task type
Story
More actions | clear | save

The purpose of the move is to put the fields people use most at the top of the form so they are the easiest to access.

frontend/src/app/tasks/hooks/useTaskTypes.ts
negreirosleo: oops, I started with useTask and forgot to change this type to TaskType, tnx

dmtrek14 (approved):
Looks good to me - thank you!

anarute (approved):

Issue #682, opened 2023-09-26
Add basic permission checking in API


dmtrek14:
We first need to make sure that the user has a user record in the database. We can introspect the token and get the username and use that to check the db for the user. If no record is found, we throw a 401. If a record is found, we fill in a user object with basic details. For some endpoints, we will need to make sure that the user_id of an object matches the id of the user we have just checked.

PR #683, opened 2023-09-27
Implement user check and start of permissions

dmtrek14:
Addresses #682 and is an initial implementation of permissions for the API.

- Add get_current_user to dependencies that introspects token and builds user info based on username in token
- Add config values to indicate use of OIDC roles (passed in token) or roles in local db
- Add the user dependency to template endpoints and check user id against logged in user
- Update user models and schema as needed
- Add user service to get users and user roles
- Move Depends(BearerToken()) up to entire timelog router so it no longer needs to be added to each endpoint

Code for the user check based on the examples shown in https://fastapi.tiangolo.com/tutorial/security/simple-oauth2/.


api/routers/v1/timelog.py
```diff
@@ -13,24 +13,32 @@
...
-@router.get("/task_types/", dependencies=[Depends(BearerToken())], response_model=List[TaskTypeItem])
+@router.get("/task_types/", response_model=List[TaskTypeItem])
 async def get_task_types(db: Session = Depends(get_db), skip: int = 0, limit: int = 100):
     items = TaskTypeService(db).get_items()
     return items
 
 
-@router.get("/templates/{user_id}", dependencies=[Depends(BearerToken())], response_model=List[TemplateSchema])
-async def get_user_templates(user_id: int, db: Session = Depends(get_db)):
+@router.get("/templates/{user_id}", response_model=List[TemplateSchema])
```
anarute, line 32:
```suggestion
@router.get("/templates/", response_model=List[TemplateSchema])
```

I know this was there before, but I think it's better to remove the user_id from this route to keep it consistent with tasks and because I don't think it makes sense to have a path like `templates/1` that gets the user 1 templates instead of the template `1`.

api/routers/v1/timelog.py
```diff
@@ -13,24 +13,32 @@
...
+async def get_user_templates(user_id: int, user=Depends(get_current_user), db: Session = Depends(get_db)):
+    if user_id is not user.id:
```
anarute, line 34: Although this might be working, the `is` and `is not` should be used only when comparing [identity](https://docs.python.org/3/reference/expressions.html#is), in this case we want to compare the values, so better to use the `!=` operator

```suggestion
    if user_id != user.id:
```

Some background on this:
https://discuss.python.org/t/demoting-the-is-operator-to-avoid-an-identity-crisis/86
https://stackoverflow.com/questions/132988/is-there-a-difference-between-and-is

In the other PRs flake8 was complaining because we were comparing to True or False directly, which is an anti pattern: https://www.flake8rules.com/rules/E712.html

api/routers/v1/timelog.py
```diff
@@ -46,18 +54,26 @@ async def add_template(template: TemplateNewSchema, db: Session = Depends(get_db
     \f
     :param item: User input.
     """
+    if template.is_global and "manager" not in user.roles:
+        raise HTTPException(status_code=403, detail="You are not authorized to create global templates")
+    else:
+        if template.user_id is not None and template.user_id is not user.id:
```
anarute, line 60: Same here:

```suggestion
        if template.user_id is not None and template.user_id != user.id:
```

and to simplify:

```suggestion
        if template.user_id and template.user_id != user.id:
```

api/routers/v1/timelog.py
```diff
@@ -46,18 +54,26 @@ async def add_template(template: TemplateNewSchema, db: Session = Depends(get_db
...
 async def get_user_tasks(
     user_id: int,
+    user=Depends(get_current_user),
     db: Session = Depends(get_db),
     offset: int = Query(0, ge=0),
     limit: int = Query(25, ge=0, le=500),
     start: date = date.today(),
     end: date = date.today(),
 ):
+    if user.id is not user_id and "manager" not in user.roles:
```
anarute, line 76:
```suggestion
    if user.id != user_id and "manager" not in user.roles:
```

api/routers/v1/timelog.py
```diff
@@ -13,24 +13,32 @@
...
+@router.get("/task_types/", response_model=List[TaskTypeItem])
 async def get_task_types(db: Session = Depends(get_db), skip: int = 0, limit: int = 100):
     items = TaskTypeService(db).get_items()
     return items
 
 
-@router.get("/templates/{user_id}", dependencies=[Depends(BearerToken())], response_model=List[TemplateSchema])
-async def get_user_templates(user_id: int, db: Session = Depends(get_db)):
+@router.get("/templates/{user_id}", response_model=List[TemplateSchema])
+async def get_user_templates(user_id: int, user=Depends(get_current_user), db: Session = Depends(get_db)):
```
anarute, line 33: To avoid confusion, I think it's better to rename `user` to `current_user`, what do you think?

api/routers/v1/timelog.py
```diff
@@ -13,24 +13,32 @@
...
+    if user_id is not user.id:
+        raise HTTPException(status_code=403, detail="You are not authorized to see templates for this user")
     templates = TemplateService(db).get_user_templates(user_id)
     return templates
 
 
-@router.post("/templates", dependencies=[Depends(BearerToken())], response_model=TemplateSchema, status_code=201)
-async def add_template(template: TemplateNewSchema, db: Session = Depends(get_db)):
+@router.post("/templates", response_model=TemplateSchema, status_code=201)
+async def add_template(template: TemplateNewSchema, user=Depends(get_current_user), db: Session = Depends(get_db)):
```
anarute, line 41: Here as well: user -> current_user

anarute (changes requested):

api/routers/v1/timelog.py
dmtrek14: That makes sense to me. Will make the change.

.env.example
```diff
@@ -25,6 +25,9 @@ OIDC_AUTHORITY=
 OIDC_CLIENT_ID=
 OIDC_USERNAME_PROPERTY=
 OIDC_CERTS_URL=
+OIDC_TOKEN_ENDPOINT=
+USE_OIDC_ROLES=false
```
anarute, line 29:
```suggestion
USE_OIDC_ROLES=
```

This is probably a limitation from the `decouple` lib but passing `false` here is being interpreted as a string and then the check `if USE_OIDC_ROLES:` is true, even when we want to have it false. So better to leave it empty if we want this to be false.

anarute (approved):
It looks good to me, thank you for the changes!

PR #685, opened 2023-10-04
Add Timer to Task Form

negreirosleo:
### Description

This PR adds the Timer to the Task Form.

### Testing

1. Click Start Timer
2. It should start counting the logged time and set the "From" select to the current time
3. Click Stop Timer
4. It should stop counting the logged time and set the "To" select to the current time
5. The selects should be clearable
6. The selects should be disabled while the timer is running
7. You can also manually edit the From/To time by selecting a specific time withing 15min intervals or just writing the time.

### Screenshots/Videos
[Screencast from 2023-10-04 16-03-00.webm](https://github.com/Igalia/phpreport/assets/20492786/2e909528-2ec7-4b7c-86c8-766a2ac5ba89)


frontend/src/ui/Select/Select.tsx
```diff
@@ -4,37 +4,83 @@ import Box from '@mui/joy/Box'
...
+        loading={loading}
+        getOptionLabel={getOptionLabel}
+        disabled={disabled}
+      />
+    </Box>
+  )
+}
 
-export const Select = ({ options, sx, onChange, name, value, label, loading }: SelectProps) => {
+export const FreeSoloSelect = ({
```
negreirosleo, line 51: I created this second `Select` because of the `freeSolo` prop (https://mui.com/joy-ui/react-autocomplete/#search-input), it adds the the possibility of writing thing that are outside of the select options, in this case it was the time select for the timer, it has options within a 15min interval but you can still add different times.

Because of the way that the Select is built, when you add the `freeSolo` prop, the type of the value becomes `T | string` instead of just `T`, so instead of having to check in the `onChange`  if the option is a string even when we know it will never be a string I decided to create a another similar component that only has string options for those cases.

dmtrek14:
Currently, we use 24h time notation instead of 12h. I am open to using either format or allowing the user to choose (or set a preference later on if that is something we want to let a user control). What do you think, @anarute? I don't have strong feelings one way or the other; I am mostly raising it since it will affect our validation for the times.

dmtrek14:
I noticed a warning and an error in the console. Are these anything to be concerned about? For the Task Type, I have selected 'mock-test' from the dropdown.
Screenshot from 2023-10-05 12-56-11

negreirosleo:
@dmtrek14 I'm aware of those warnings, I will remove the JoyUI related warnings after I finish the form test

dmtrek14:
When I clear the form with the Clear button, neither of the dropdowns are clearing (Project, Task Type). This happens whether I use the timer, select the time, or input it.

anarute:
> Currently, we use 24h time notation instead of 12h. I am open to using either format or allowing the user to choose (or set a preference later on if that is something we want to let a user control). What do you think, @anarute? I don't have strong feelings one way or the other; I am mostly raising it since it will affect our validation for the times.

Let's keep the 24h format and think about allowing the user to choose in the future

anarute:
The timer looks good to me, there are a few things I think we need to improve for the form validation, but it can be in a future PR for validation, I'll create an issue.

anarute (approved):

dmtrek14 (approved):
LGTM

PR #686, opened 2023-10-04
Add PUT template endpoint

### LINKED ISSUE #681: Add PUT endpoint for templates

dmtrek14:
Closes #681 

- Fix regression where template model validator was lost in upgrade to Pydantic
- Add endpoint and basic checks: same user checks as for adding a template, check that template user is seeking to edit exists
- Add simple schema for update
- Add service methods to get single template and to update
- Add additional validation for templates to check if there is a task_type and if there is to make sure it exists (otherwise there is a FK error)

### DIFF
--- a/api/routers/v1/timelog.py
+++ b/api/routers/v1/timelog.py
@@ -7,16 +7,17 @@ from schemas.timelog import (
     TaskTypeItem,
     Template as TemplateSchema,
     TemplateNew as TemplateNewSchema,
+    TemplateUpdate as TemplateUpdateSchema,
     Task as TaskSchema,
     TaskNew as TaskNewSchema,
-    TaskValidate,
 )
+from schemas.validation import ValidatedObject
 from services.timelog import TaskTypeService, TemplateService, TaskService
 from services.projects import ProjectService
 from services.config import ConfigService
 from db.db_connection import get_db
 from auth.auth_bearer import BearerToken
-from dependencies import get_current_user
+from dependencies import get_current_user, AppUser
 
 router = APIRouter(
     prefix="/timelog",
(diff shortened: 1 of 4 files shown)


api/routers/v1/timelog.py
```diff
@@ -66,10 +67,52 @@ async def add_template(
...
             raise HTTPException(status_code=403, detail="You are not authorized to create templates for this user")
+    if template.task_type:
+        task_type_valid = TaskTypeService(db).slug_is_valid(template.task_type)
+        if not task_type_valid:
+            raise HTTPException(status_code=422, detail=f"Task type {template.task_type} does not exist.")
     result = TemplateService(db).create_template(template)
     return result
 
 
+@router.put("/templates", response_model=TemplateSchema)
```
anarute, line 78: For this case I think it would be nice to have the template_id in the url:

```suggestion
@router.put("/templates/{template_id}", response_model=TemplateSchema)
```

api/routers/v1/timelog.py
```diff
@@ -66,10 +67,52 @@ async def add_template(
...
+        raise HTTPException(status_code=404, detail=f"Template with id {template_id} not found")
+    if template.is_global and "manager" not in current_user.roles:
+        raise HTTPException(status_code=403, detail="You are not authorized to update global templates")
+    else:
+        if template.user_id and template.user_id != current_user.id:
+            raise HTTPException(status_code=403, detail="You are not authorized to update templates for this user")
+    if template.task_type:
+        task_type_valid = TaskTypeService(db).slug_is_valid(template.task_type)
+        if not task_type_valid:
+            raise HTTPException(status_code=422, detail=f"Task type {template.task_type} does not exist.")
```
anarute, line 111: This is the same logic from creating a template, should we move it to a method?

dmtrek14: I moved it to its own method and did a small refactor: I had a TaskValidate object, but moved it to validation.py and repurposed it as a general validator object that we could use for anything.

anarute (approved):
Looks good to me, thanks!

PR #690, opened 2023-10-10
Add users/me endpoint

anarute:
- Move the AppUser to user schemas
- Add `users/me` endpoint to fetch more details about the current user, incluing the user_id that is needed in order to use other endpoints


dmtrek14 (approved):
Looks good to me - thanks!

PR #693, opened 2023-10-10
Add PUT endpoint for tasks

### LINKED ISSUE #692: Add PUT endpoint for tasks

dmtrek14:
Closes #692 

- Add the endpoint
- Add TemplateUpdate schema
- Add a simple get method to tasks service so we can check for existing task
- Small updates to endpoint docs to provide more clarity
- Update validation.

The validation updates for update were a bit extensive. Since it should be possible for someone to send a PUT request with only the task_id and the properties they want to update, we needed several additional checks to ensure the validation would work with a partial object. Mostly, we need to ensure the required properties (user_id, project_id, end, init, date) are on the object to be validated or else the validation fails.

### DIFF
--- a/api/routers/v1/timelog.py
+++ b/api/routers/v1/timelog.py
@@ -1,6 +1,7 @@
 from typing import List
 from datetime import date
 from fastapi import APIRouter, Depends, Query, HTTPException
+from fastapi.encoders import jsonable_encoder
 from sqlalchemy.orm import Session
 
 from schemas.timelog import (
@@ -10,6 +11,7 @@ from schemas.timelog import (
     TemplateUpdate as TemplateUpdateSchema,
     Task as TaskSchema,
     TaskNew as TaskNewSchema,
+    TaskUpdate as TaskUpdateSchema,
 )
 from schemas.validation import ValidatedObject
 from services.timelog import TaskTypeService, TemplateService, TaskService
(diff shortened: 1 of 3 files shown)


dmtrek14:
@anarute I put this one in draft status to get some feedback. The validate_task method got lengthier here and I think I could use another pair of eyes. Instead of a full object merge, I decided to validate an updated task only if the properties we validate (start, end, project_id, date) are different from the task in the db. If someone only wants to update the description or story, is it necessary to revalidate everything? I went with 'no' but am open to going the other way.

api/routers/v1/timelog.py
```diff
@@ -143,25 +145,95 @@ async def add_task(task: TaskNewSchema, current_user=Depends(get_current_user),
...
+    :param item: User input.
+    """
+    existing_task = TaskService(db).get_task(task_id)
+    if not existing_task:
+        raise HTTPException(status_code=404, detail=f"Task with id {task_id} not found")
+    user_id_to_check = task.user_id if task.user_id else existing_task.user_id
+    if current_user.id != user_id_to_check:
+        raise HTTPException(status_code=403, detail="You are not authorized to update tasks for this user.")
+    if not task.user_id:
+        task.user_id = user_id_to_check
```
anarute, line 174: This check is been done twice, this is minor but can simplify a bit with a reorder:

```suggestion
    if not task.user_id:
        task.user_id = existing_task.user_id
    if current_user.id != task.user_id:
        raise HTTPException(status_code=403, detail="You are not authorized to update tasks for this user.")
```

dmtrek14: Since I changed the update to simply merge the changes with the existing task, this is probably no longer relevant.

anarute (approved):
That's much better, thank you!

PR #697, opened 2023-10-11
Add minimal setup for the API tests

anarute:
Initially it only tests the status endpoint, but I'll start creating tests for the existing services and endpoints.

To run the tests locally you can do it through tox: `tox -e pytest` or running `pytest -v` directly from the api folder.


dmtrek14 (approved):
Looks good to me - thanks!

PR #700, opened 2023-10-11
Add Task Form Submission

negreirosleo:
### Description

This PR creates the Task Form submission, and adds an Alert component to indicate if the an action is successful or not.

### Testing

1. Go to `/web/v2/tasks`
2. Select a project
3. Select the time
4. Fill other fields (optional)
5. Click `Save`
6. The form should submit and a Success or Error alert should appear
7. Pressing Ctrl+S should do the same action as clicking Save
8. The Alert should close after 15s
9. Clicking `Close` also should close the alert

### Screenshots

[Screencast from 2023-10-11 17-09-50.webm](https://github.com/Igalia/phpreport/assets/20492786/ecd78e66-d5cc-42e9-b7b0-bb2f84cad102)

### DIFF
--- a/frontend/src/app/providers.tsx
+++ b/frontend/src/app/providers.tsx
@@ -4,6 +4,7 @@ import { CssVarsProvider } from '@mui/joy/styles'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
 import { AuthProvider } from '@/app/auth/AuthProvider'
 import { theme } from '@/ui/theme'
+import { AlertProvider } from '@/ui/Alert/AlertProvider'
 
 const queryClient = new QueryClient()
 
@@ -11,7 +12,9 @@ export const Providers = ({ children }: { children: React.ReactNode }) => {
   return (
     <AuthProvider>
       <QueryClientProvider client={queryClient}>
-        <CssVarsProvider theme={theme}>{children}</CssVarsProvider>
+        <AlertProvider>
+          <CssVarsProvider theme={theme}>{children}</CssVarsProvider>
+        </AlertProvider>
       </QueryClientProvider>
     </AuthProvider>
   )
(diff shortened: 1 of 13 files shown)


anarute:
Thank you for this patch! A few comments:

- Can we focus to the form once the page is loaded? this is to make the form submission from the keyboard easier - unless this breaks the accessibility somehow @dmtrek14?
- The time fields should have a mask for HH:MM, so when entering 1000 for instance it's converted to 10:00. In case it's less than 3 digits, it should add a 0 in front, so if the person enters 915, it should convert to 09:15. In case the user inputs 2 digits, 2 zeros should be added, so 12 becomes 12:00
- When pressing enter I expect the form is submitted but I'm taken to the Project input instead

dmtrek14:
> * Can we focus to the form once the page is loaded? this is to make the form submission from the keyboard easier - unless this breaks the accessibility somehow @dmtrek14?

I will check, but I think this could be a violation. A page should be navigable sequentially and setting the focus on the form changes that a little. The 'skip navigation' is helpful here in that a user can tab once when hitting the tasks page, enter 'skip navigation' and the focus will then be on the form.

dmtrek14:
In the case of the mask, I will just note that there is a regex on the model validation that will take 1 or 2 digits for the hour (it isn't in this branch, though), so the backend doesn't care so much if it is sent '09:15' or '9:15'. We just need the ':' separating the values. I think it is fine if we want to sanitize the inputs on the front, for consistency's sake, though.

I also expected 'enter' to submit the form. 

In the existing app, we also auto-save after 30 seconds (provided input is valid), so that is probably something to factor in here.

anarute (approved):

dmtrek14 (approved):
All of this is working well for me. I think we can continue to refine the mask as we move forward.

PR #702, opened 2023-10-12
Increase login column size for username

anarute:
The migration was created already on 35830895 but we missed updating the model.


dmtrek14 (approved):
LGTM - good catch.

PR #705, opened 2023-10-17
Remove duplicate 'users' in route

dmtrek14:


anarute (approved):
Thanks for catching it!

PR #707, opened 2023-10-17
Add next-auth

negreirosleo:
### Description:

This PR:
- replaces the `react-oidc-context` lib with `next-auth` to handle the authentication
- adds a middleware for protected routes.
- replaces `axios` for `fetch` to [automatically cache the data](https://nextjs.org/docs/app/building-your-application/data-fetching/fetching-caching-and-revalidating#caching-data)
- creates a sign-in page, but its temporary and susceptible to further changes

### Testing:

1. Try to navigate to `/tasks` without logging in.
2. It should redirect you to the sign-in page
3. After the sign-in it should redirect you to the home page `/`
4. Acessing `/task` now should be possible


frontend/src/app/auth/AuthorizedPage.tsx
dmtrek14, line 1: I like the move of checking authentication into middleware, but I think we will still need this component. I would like to repurpose it to check a user's roles for a page - add a param called `authorizedRoles` that takes an Array<string> of roles allowed to view the page. 

I suppose the check for roles could be done in the middleware, too, but I would be afraid of bloating the middleware. What do you think?

frontend/src/app/tasks/page.tsx
dmtrek14, line 1: If you agree with leaving the <AuthorizedPage> component to check user roles, then we would want to leave that component in here.

frontend/src/infra/project/getProjects.ts
```diff
@@ -0,0 +1,12 @@
+import { apiClient } from '@/infra/lib/apiClient'
+import { Project } from '@/domain/Project'
+
+export const getProjects = async (): Promise<Array<Project>> => {
+  const response = await apiClient('/v1/projects/')
+
+  if (!response.ok) {
```
dmtrek14, line 7: Is the plan to bubble up error messages to the user as we evolve this code?

frontend/src/middleware.ts
```diff
@@ -0,0 +1,27 @@
+import { getToken } from 'next-auth/jwt'
+import { NextResponse } from 'next/server'
+import type { NextRequest } from 'next/server'
+import { validateToken } from './infra/lib/validateToken'
+
+const protectedRoutes = ['/tasks']
```
dmtrek14, line 6: Eventually, all routes will be protected except for maybe a generic landing page. Nothing to change here - just making that note.

dmtrek14 (commented):
The only thing I had a question about is with regard to the <AuthorizedPage> component and the middleware, and then I know you are working on a small patch for the middleware. Otherwise, this is a great start that we can build on. Thanks for this!

frontend/src/infra/project/getProjects.ts
negreirosleo: Yes! thats the plan, this throw is just a placeholder for now

dmtrek14 (commented):
With the adjustments to the middleware, I am not redirected to the keycloak signin. When I navigate to the Tasks page, I am given the 'Failed to fetch projects' error. I tried making sure that I was already signed out of the OAuth provider. I also tried when I was already signed in.

negreirosleo:
I will add the configuration to redirect, thanks Danielle! I thought the next-auth lib was doing it but I checked here and got the same thing

PR #708, opened 2023-10-18
Improve tests setup and add get task type test

anarute:
- add token creation logic and use it in the test setup
- add initial mock data
- add get task types test


docker/docker-compose.test.yml
dmtrek14, line 1: My api container doesn't come up. I am getting the following error in the log:
Screenshot from 2023-10-18 10-08-49

PR #713, opened 2023-10-20
Add test coverage to the API

anarute:
Run pytest with coverage and add a coverage report step in tox.

Next step would be adding a badge to the project, but for now I think this should be enough.


anarute:
The result report: https://github.com/Igalia/phpreport/actions/runs/6587407102/job/17897631495?pr=713#step:8:39

dmtrek14 (approved):
LGTM - thanks for adding!

PR #714, opened 2023-10-24
Add basic permissions

dmtrek14:
The middleware is doing the check to make sure someone is logged in and the getCurrentUser is making sure the user is in the local database. 

This adds a basic start for permissions/authorization in the app, which is primarily located in the useIsAuthorized hook. This does a simple check to see if the user's roles contain the authorized roles; right now, it is only implemented for the page level on the tasks page and any role is allowed there. The hook could be used for permission checking on individual components in the future. 

To test, make sure that your local user in the phpreport database belongs to one of the following user groups:
- Staff
- Manager
- Admin
- Human resources
- Project manager

To add these groups to an existing database in docker, run:
```
docker exec -ti phpreport-db bash
psql phpreport -U phpreport
INSERT into user_group(id, name) VALUES (DEFAULT, 'human resources'), (DEFAULT, 'project manager');
```
To add your user to a group, run 
`INSERT into belongs VALUES({user group id of group you want to be added to}, {your user id});`

As long as you have one of these roles, you will be able to access the tasks page. If you remove them, you should get thrown to an error page telling you that you are not authorized to see the page.

Make sure to also run the migration that is part of this PR - it adds a column to the user_group table to hold scopes and populates the default values.


anarute:
Not related to your changes directly, but we should add new lines in the end of the files - this can be a setting in your code editor so you don't need to remember to do it.

image


A few reasons why: https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

frontend/src/app/tasks/page.tsx
```diff
@@ -14,8 +15,14 @@ const getPageData = async () => {
   ])
 }
 
+const authorizedRoles = [ "staff", "manager", "admin", "project manager", "human resources"]
```
anarute, line 18: In order not to have this hardcoded, how would we solve it? I vaguely recall discussing this with you, would it be creating a map of routes and roles and storing in the db?

anarute (approved):
Seems like a good start to start managing the permissions, thank you!

frontend/src/app/tasks/page.tsx
dmtrek14: I am still thinking about this part. I am leaning towards something like what you mention, but I have also seen applications put something like this in the middleware - with hardcoded strings or an enum. I was initially worried about bloating the middleware, but it could be worth exploring doing the check there. 

I have a couple of other changes I think I am going to make before pulling this out of draft status, so I can experiment a little.

anarute: The problem is that we need this check in the backend too so in any case we need them stored somewhere no only in the middleware.

I'll try to do some research too for solutions that don't over complicated.

frontend/src/hooks/useIsAuthorized/useIsAuthorized.ts
```diff
@@ -0,0 +1,13 @@
...
+    if(currentUser){
+        if(!currentUser.roles){
+            return false;
+        }
+        return currentUser.roles.some((role) => authorizedRoles.includes(role));
+    }
+    else {
+        return false;
+    }
+}
```
negreirosleo, line 13: How about:

```ts
export const isAuthorized = (currentUser: User, authorizedRoles: Array<string>) => {
  if(!currentUser?.roles) return false
  
  return currentUser.roles.some((role) => authorizedRoles.includes(role))
}
```

or
```ts
export const isAuthorized = (role: string, authorizedRoles: Array<string>) => {  
  roles.some((role) => authorizedRoles.includes(role))
}
```

Also, this doesn't need the `use` prefix because technically it isn't a hook, just a regular js function.
Check this on the `deep dive` section: https://react.dev/learn/reusing-logic-with-custom-hooks#hook-names-always-start-with-use

frontend/src/app/auth/AuthorizedPage.tsx
```diff
@@ -0,0 +1,14 @@
+'use client'
+import { User } from "@/domain/User";
+import { useIsAuthorized } from "@/hooks/useIsAuthorized/useIsAuthorized";
+
+export const AuthorizedPage = ({ children, currentUser, authorizedRoles }: { children: React.ReactNode, currentUser: User, authorizedRoles: Array<string> }) => {
```
negreirosleo, line 5: I think here we could only send the `roles` and not the whole `currentUser`

frontend/src/hooks/useIsAuthorized/useIsAuthorized.ts
dmtrek14: I actually meant to delete this hook entirely now that I moved the first check to the middleware. For protecting specific components on a page - like the new 'delete task ' button Leonardo has added - it is a simple check for a scope (`user.authorizedScopes.includes("task:delete-own")`) and I don't think that this needs a separate function now.

api/migrations/versions/72e6af46be22_add_scope_column_to_user_group_table.py
```diff
@@ -0,0 +1,73 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("user_group", sa.Column("scopes", sa.String, nullable=True))
+    # Add scopes for default roles
+    separator = ","
+    scopes_everyone_has = "task:create-own,task:read-own,task:update-own,task:delete-own,task_type:read,template:create-own,template:update-own,template:read-own,template:read-global,template:delete-own,project:read-assigned,vacation:read-self,vacation:read-other,vacation:create-own,vacation:update-own,vacation:delete-own"
```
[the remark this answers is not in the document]
dmtrek14: Agreed. This was probably a case of tired eyes. Will fix.

frontend/src/app/auth/error/page.tsx
```diff
@@ -0,0 +1,15 @@
+'use client'
+
+type ErrorProps = {
+    error: Error;
+}
+export default function Error({error}: ErrorProps) {
+    console.log(error)
```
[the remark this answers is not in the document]
dmtrek14: 🤦🏻

api/migrations/versions/72e6af46be22_add_scope_column_to_user_group_table.py
```diff
@@ -0,0 +1,73 @@
...
+        f"UPDATE user_group SET scopes = '{hr_scopes}' WHERE name = 'human resources'"
+    )
+    conn.execute(hr_query)
+    manager_query = sa.text(
+        f"UPDATE user_group SET scopes = '{manager_scopes}' WHERE name = 'manager'"
+    )
+    conn.execute(manager_query)
+    admin_query = sa.text(
+        f"UPDATE user_group SET scopes = '{admin_scopes}' WHERE name = 'admin'"
+    )
```
anarute, line 64: These groups don't exist yet, so maybe we need to create them in this migration as well?

dmtrek14: Good point. I think we have a note in the dev docs about how to add some groups with a sql statement, but nothing in any of the migrations. I will add this to the migration.

anarute (approved):
I've tested locally and seems to be working fine, looks good to me, thanks!

frontend/src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -9,20 +10,41 @@ export const authOptions: NextAuthOptions = {
...
       return `${baseUrl}/web/v2/`
     },
     async session({ session, token }) {
+      const headers = {
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${token.accessToken}`
+      }
+
+      const response = await fetch(`${process.env.API_BASE}/v1/users/me`, { headers: { ...headers } })
+      const result = await response.json()
```
Comment (negreirosleo) [inline frontend/src/app/api/auth/[...nextauth]/route.ts:27]: we might not need this here, we can use `token.user` since we already fetched it on the `jwt` method:

image

frontend/src/middleware.ts
```diff
@@ -22,6 +22,17 @@ export async function middleware(request: NextRequest) {
...
+    if(!session.user.id){
+      return new NextResponse("You do not have a user record in the application. Please contact your sysadmin.", {status: 401})
+    }
+    if(!session.user.roles || session.user.roles.length ==0){
+      return new NextResponse("You have not been assigned any roles in the application. Please contact your sysadmin.", {status: 403})
+    }
+    if(path == "/tasks" && !session?.user.authorizedScopes.includes("task:create-own")){
+      return NextResponse.redirect(new URL('/web/v2/auth/error?error=AccessDenied', request.url))
+    }
+  }
```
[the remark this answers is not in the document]
negreirosleo: Also I made a mistake in this file and just noticed it, 

```ts
  const session = await getToken({
    req: request,
    secret: process.env.NEXTAUTH_SECRET
  })
```
`getToken` doesn't return a session, so we should rename this from `session` to `token`, can we fix it in this PR?

frontend/src/app/api/auth/[...nextauth]/route.ts
Comment (dmtrek14) [reply frontend/src/app/api/auth/[...nextauth]/route.ts]: I think you are right. I had initially added the fetch to the session callback and was not getting the data at the right time and must have forgotten to remove it after I had it working. Thanks for the catch.

frontend/src/middleware.ts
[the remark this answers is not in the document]
dmtrek14: Definitely. Just pushed up the fix.

negreirosleo (approved):
Tested here and it works, the code LGTM, thanks for the changes!

PR #715, opened 2023-10-24
Default env variables for tests to None

dmtrek14:
It seems that `decouple` doesn't resolve a not-found .env variable to None by default. Since the JWT_PUB_KEY and JWT_PRIVATE_KEY are here just for testing, we should probably force the default to None so that `decouple` doesn't toss errors when running for development.


anarute (approved):
Nice catch, thank you!

PR #717, opened 2023-10-25
Split front and back tests in GH actions

anarute:
The frontend tests don't need to run on docker nor need the DB running so better to split them to get an easier to read output.


dmtrek14 (approved):
LGTM - thanks!

PR #718, opened 2023-10-27
Add set of tests for Tasks endpoints

anarute:
- minor fixes in the task schemas
- add tests for GET, CREATE, PUT and DELETE endpoints


dmtrek14 (approved):
Everything here works for me as expected. Thanks for sorting out the models so they don't conflict with the tests.

PR #719, opened 2023-11-01
Add user capacity to AppUser object

dmtrek14:
This adds the 'capacities' list to the User object in front and back end. A capacity is defined here as the number of hours a user works per week for the provided timeframe. We use the capacity to do several of the calculations in the summary, so it makes sense to include it as part of the user payload.


anarute (approved):
LGTM, thank you!

PR #720, opened 2023-11-02
Patch login service

dmtrek14:
- Make oidc client init method public so it can be used elsewhere
- Add check for use of external auth. If using, make a curl request to the token endpoint to log user in with OIDC, set access token and user.

The loginService.php is only used with the old API and it needs to bypass the OAuth provider login page and redirect. This is why the auth request here is made via curl to the token endpoint.


anarute (approved):
LGTM, thanks!

PR #721, opened 2023-11-08
Add task overlap check

negreirosleo:
Adds overlap check to the task form, if a task is overlapping with another it won't submit the form.

[Screencast from 2023-11-08 18-03-44.webm](https://github.com/Igalia/phpreport/assets/20492786/67c8d419-0377-4db9-8deb-e8bf90cee555)

### DIFF
--- a/frontend/src/app/tasks/__tests__/TaskForm.test.tsx
+++ b/frontend/src/app/tasks/__tests__/TaskForm.test.tsx
@@ -9,8 +9,6 @@ const setupTaskForm = () => {
     {
       id: 1,
       is_active: true,
-      init: null,
-      end: null,
       invoice: null,
       estimated_hours: null,
       moved_hours: null,
@@ -350,8 +348,6 @@ describe('TasksPage', () => {
         startTime: '12:12',
         endTime: '13:14',
         id: 18,
-        init: 672,
-        end: 674,
         projectName: 'Holidays',
         customerName: 'Internal'
       }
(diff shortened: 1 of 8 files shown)


frontend/src/infra/task/getTasks.ts
```diff
@@ -0,0 +1,19 @@
...
+  const params = new URLSearchParams({ user_id: userId.toString(), start, end })
+  const response = await apiClient(`/v1/timelog/tasks/?${params}`)
+
+  if (!response.ok) {
+    throw new Error('Failed to fetch Tasks')
+  }
+
+  const data = await response.json()
+
+  return camelizeKeys(data)
```
anarute, line 18: Why do we need this method?

frontend/src/app/tasks/__tests__/TaskForm.test.tsx
```diff
@@ -281,4 +335,41 @@ describe('TasksPage', () => {
       userId: 0
     })
   })
+
+  it("doesn't submit if it have overlapping tasks", async () => {
```
anarute, line 339: typo

```suggestion
  it("doesn't submit if it has overlapping tasks", async () => {
```

anarute (approved):
It would be great if the commits were more atomic and the messages address a bit why some changes are made, not just what was done, it helps to review and in case we need to revert something in the future we will have more context.

Other than that it looks good to me, thank you!

frontend/src/app/tasks/__tests__/TaskForm.test.tsx
negreirosleo: oops, I will fix it thanks!

frontend/src/infra/task/getTasks.ts
negreirosleo: because the api sends the json properties in snake_case, this transforms to camelCase

```
{
  project_id: 0
}

// becames

{
  projectId: 0
}
```

anarute: Isn't it a bit overhead? We can send from the API in camel case already if that's better

PR #722, opened 2023-11-10
Update schemas to return camel case json

dmtrek14:
- Use alias generator from pydantic to emit camel case and set populate_by_name to true so validation does not fail
- Update fill of capacities list in get_current_user due to model type mismatch after update of schemas
- Update json in tests to match new schema

Regarding the second item here: prior to changing the model_config for UserCapacity, it seems that the mapping from model.user.UserCapacity => schemas.user.UserCapacity happened under the hood easily. After the change, pydantic was quite unhappy about the type mismatch, so I did a simple iterate and append. If I missed some easier way to address this, please let me know.

This was the error received after the model config change:
```
File "/usr/local/lib/python3.9/dist-packages/pydantic/main.py", line 164, in __init__
2023-11-10 11:02:13     __pydantic_self__.__pydantic_validator__.validate_python(data, self_instance=__pydantic_self__)
2023-11-10 11:02:13 pydantic_core._pydantic_core.ValidationError: 1 validation error for AppUser
2023-11-10 11:02:13 capacities.0
2023-11-10 11:02:13   Input should be a valid dictionary or instance of UserCapacity [type=model_type, input_value=<models.user.UserCapacity...bject at 0x7f82fb06d430>, input_type=UserCapacity]
2023-11-10 11:02:13     For further information visit https://errors.pydantic.dev/2.4/v/model_type
```

### DIFF
--- a/api/dependencies.py
+++ b/api/dependencies.py
@@ -5,7 +5,7 @@ from fastapi.security import OAuth2PasswordBearer
 from decouple import config
 from db.db_connection import get_db
 from auth.auth_handler import decode_token
-from schemas.user import AppUser
+from schemas.user import AppUser, UserCapacity
 from services.user import UserService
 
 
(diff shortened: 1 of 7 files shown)


anarute (approved):
LGTM, thanks!

PR #727, opened 2023-11-14
Add task deletion to task list

negreirosleo:
Clicking on the trashcan icon will call the task deletion endpoin

### Screenshots

[Screencast from 2023-11-14 13-18-43.webm](https://github.com/Igalia/phpreport/assets/20492786/9a95a1b3-88cc-428e-88fb-effaaeef8b90)

### DIFF
--- a/frontend/src/app/tasks/TaskList.tsx
+++ b/frontend/src/app/tasks/TaskList.tsx
@@ -1,9 +1,12 @@
 'use client'
 
 import Box from '@mui/joy/Box'
+import Button from '@mui/joy/Button'
 import Typography from '@mui/joy/Typography'
-
-import { useGetTasks } from './hooks/useTask'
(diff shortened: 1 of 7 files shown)


anarute (approved):
LGTM, thanks!

dmtrek14 (commented):
@anarute Do we want there to be a confirmation for the deletion? The current app doesn't have a confirmation, but someone does have to delete, then save changes. So it could be nice to make sure someone really wants to delete.

anarute:
> then save changes.

Not really because the current version has the autosave feature.  But agree we can add it as there's no undo.

dmtrek14 (approved):
Looks good here - thanks for the confirmation on delete!

anarute (approved):
Thank you for the changes!

PR #730, opened 2023-11-17
Create user session hook

negreirosleo:
This PR creates the `useGetCurrentUser` hook, it returns the user data from the session. In the route we add the user data to the session and now it can be accessed at any client component trough this hook.


dmtrek14 (approved):
LGTM - thanks for this addition.

PR #731, opened 2023-11-17
Fix Project type

negreirosleo:
The type was written with snake_case but it should be camelCase to fit with the project patterns


dmtrek14 (approved):
LGTM

PR #733, opened 2023-11-17
Implement refresh access token logic

dmtrek14:
Add refresh token logic

- Add function to the auth route to check the expiry of the token and attempt a refresh against the OIDC provider endpoint
- Add necessary fields to next-auth types


negreirosleo (approved):
it's working for me :+1:

anarute (approved):

PR #735, opened 2023-11-30
Add totp param to passthrough to oauth

dmtrek14:
- Add totp so that users with 2FA will be able to use this service

A previous patch added a cURL request to the external auth provider, but did not take into account 2FA. This patch fixes that issue by accepting a 'totp' param in the url which can be passed to the OAuth provider.


anarute (approved):
LGTM

PR #737, opened 2023-12-05
Fix: Set used hours to 0 if the date is in the future

dmtrek14:
The holiday summary for an upcoming year was improperly calculating and showing some hours as used, which should not be possible. This solution checks the referenceDate passed to the method doing the calculation and if it is in the future, the used hours are set to 0.


anarute (approved):
This solution solves the problem but I wonder if there might be a hidden bug in how the addhours are being calculated and we don't know. I'm happy to merge this for now, but something to keep in mind when porting it to the new API.

PR #738, opened 2023-12-06
Fix useGetCurrentUser breaking on server

negreirosleo:
If session data doesn't exist, it only returns an empty object


dmtrek14 (approved):
LGTM - thank you!

PR #740, opened 2023-12-06
Fix select being on task form

negreirosleo:
The select was being cropped because the <body> tag on html was shorter than the screen, now it fills the whole screen but can be expanded if necessary.

image


dmtrek14 (approved):
LGTM - thanks!

PR #742, opened 2023-12-11
Fix task types dropdown

anarute:
There were 2 issues in our task types dropdown:

- when typing in the field, the autocomplete was displaying inactive task types as if they were still active.
- if a user had an inactive task type saved as a template, they could still use it.

This PR adds a filter in the endpoint that allows us to get only the active task types and it adds a check to validate the task type when a user uses a template.

### DIFF
--- a/api/routers/v1/timelog.py
+++ b/api/routers/v1/timelog.py
@@ -35,17 +35,18 @@ router = APIRouter(
 
 
 @router.get(
-    "/task_types/",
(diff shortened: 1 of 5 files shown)


dmtrek14 (approved):
Looks good to me - thanks for the patch

anarute:
Oh noes, I forgot to rebase the fixup commit before merging, sorry about that :cry:

PR #743, opened 2023-12-12
Use some instead of includes to check task types

anarute:
Includes only works for flat arrays, we want the method `some` in this case.


dmtrek14 (approved):
LGTM

PR #744, opened 2023-12-12
Create Edit Form

negreirosleo:
### Description

Add a form to the task listing that allows the user to edit the Task on the spot.

### Screenshots

[Screencast from 2023-12-12 11-31-23.webm](https://github.com/Igalia/phpreport/assets/20492786/71ce03ea-cae3-4cea-a909-f3414abc9bd3)


frontend/src/app/tasks/components/EditTask.tsx
```diff
@@ -0,0 +1,115 @@
+import Box from '@mui/joy/Box'
+import Stack from '@mui/joy/Stack'
+
+import { Task } from '@/domain/Task'
+import { TaskType } from '@/domain/TaskType'
+import { Project } from '@/domain/Project'
+import Button from '@mui/joy/Button'
```
anarute, line 7: This is very minor, but can we move the button nearer the other joy components? just to keep the imports more organized

frontend/src/app/tasks/components/EditTask.tsx
```diff
@@ -0,0 +1,115 @@
...
+type EditTaskProps = {
+  task: Task
+  tasks: Array<Task>
+  projects: Array<Project>
+  taskTypes: Array<TaskType>
+  closeForm: () => void
+}
+
+export const EditTask = ({ task, tasks, projects, taskTypes, closeForm }: EditTaskProps) => {
+  const { handleChange, formState, handleSubmit, resetForm } = useEditTaskForm({
```
anarute, line 25: This is also a question in order to learn better the architecture, why do we need hooks and not simple functions for the CRUD methods?

anarute: And also why do we have the useEditTaskForm and useEditTask? Sometimes it feels we have too many layers or things are starting to get too over engineered but it could be lack of knowledge from my side.

frontend/src/app/tasks/hooks/useTask.ts
```diff
@@ -61,3 +62,31 @@ export const useDeleteTask = () => {
...
+  const { mutate } = useMutation((task: Task) => editTask(task, apiClient), {
+    onSuccess: (data) => {
+      queryClient.setQueryData<Array<Task>>(['tasks', userId], (prevData) =>
+        prevData!.map((task) => {
+          if (task.id === data.id) {
+            return data
+          }
+
+          return task
+        })
```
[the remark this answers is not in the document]
negreirosleo: it updates the data from react query, eg:
```ts
const data = [
  {id: 1, description "task1"},
  {id: 2, description "task2"},
  {id: 3, description "task3"}
]

const updatedTask = {id: 1, description: "edited task 1"}

const updatedData = data.map(task => {
  if (task.id === updatedTask.id) {
    // here we replace the old task with the updated one
    return updatedTask
  }
  
  // leave all the other tasks alone
  return task
})

/*
Now updatedData equals to: [
  {id: 1, description "edited task 1"},
  {id: 2, description "task2"},
  {id: 3, description "task3"}
]

*/
```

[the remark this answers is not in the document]
anarute: Oh, I see thanks!

frontend/src/app/tasks/components/EditTask.tsx
anarute: ok thanks!

frontend/src/app/tasks/components/TaskBox.tsx
```diff
@@ -4,21 +4,32 @@ import { useState } from 'react'
...
 import { Task } from '@/domain/Task'
 import { getTimeDifference, convertTimeToMinutes } from '../utils/time'
 
+import { EditTask } from './EditTask'
+import { useDeleteTask } from '../hooks/useTask'
+
 type TaskProps = {
   task: Task
-  deleteTask: (taskId: number) => void
+  tasks: Array<Task>
```
dmtrek14, line 23: If this component is meant to be a single task, then why are we passing in the whole array of tasks? Is it so we can check overlap with other tasks?

frontend/src/app/tasks/page.tsx
```diff
@@ -27,7 +27,7 @@ export default async function Tasks() {
       projects={projects}
       taskTypes={taskTypes}
       templates={templates}
-      taskList={<TaskList />}
+      taskList={<TaskList projects={projects} taskTypes={taskTypes} />}
```
dmtrek14, line 30: I have an architectural question here, or it could be a confusion of names. Since a form would be concerned with just one item, why pass the list through? Would it make more sense to have a DayView outer component/container, then a form on the left that gets only the data it needs and a list component on the right that gets the list? 

My overall concern is that when we start building the week/month views, we could find our current structure not as flexible or composable as we need it to be.

dmtrek14 (commented):
Like Ana, I had a few structural questions here that might simply be my own lack of knowledge about Next.js project architecture. 

Functionally, the edit form works well so far and I like it better than the original modal concept.

frontend/src/app/tasks/components/TaskBox.tsx
negreirosleo: Yes! its basically for this function:
```ts

() => {
    const validation = Task.safeParse(formState)

    if (!validation.success) {
      validation.error.issues.map(({ message }) => {
        showError(message)
      })

      return
    }

    const { message } = getOverlappingTasks(
      formState,
      tasks.filter(({ id }) => id !== formState.id)
    )

    if (message.length > 0) {
      showError(message)
      return
    }

    editTask(formState)
```

frontend/src/app/tasks/page.tsx
negreirosleo: I think that naming of the component could be improved because it has 4 elements: Select Template, Start Timer button, the Create Task Form and the TaskList, since the first 3 all interact with the form I decided to make the TaskList just a prop so it could be tested in isolation from the form.

the current html structure is like this:
```tsx
<Container display="flex">
  <div>
    <SelectTemplate>
    <StartTimer BUtton>
  </div>
  
  <Divider>
  
  <div>
    <NewTaskForm>
    <TaskList>
  </div>
</Container>
```

We have 3 sections there: 1 top bar that interacts with the form, 2 a divider, 3 the new task form and the task list.  but I think its possible to solve this with a grid layout and make something like this:


But I think its possible to make something like:

```ts
<Container display="grid">
  <SelectTemplate>
  <StartTimerButton>
  <Divider>
  <NewTaskForm>
  <TaskList>
</Container>
```

and then we can wrapp the template select, start timer button and new TaskForm into 1 component and have something like this in the end:

```tsx
const DayView = () => {

  return (
    <Container display="grid">
      <CreateTaskForm />
      
      <Divider gridArea="middle" />
      
      <TaskList gridArea="bottom-right" />
    </Container>
  )
}

const CreateTaskForm = () => {
  return (
    <>
      <SelectTemplate gridArea="top-left" />
      <StartTimeButton gridArea="top-right" />
      <form gridArea="bottom-left" />
    </>
}
```

This way every component would be on its own place :thinking:

negreirosleo: Thanks for pointing it out @dmtrek14! I will improve that structure

negreirosleo: as soon as I finish this PR I will start that restructure

dmtrek14: Sounds good @negreirosleo - thanks for indulging me!

dmtrek14 (approved):
This looks good. Thanks for also working on the restructuring for DayView/TaskForm so that the code is more clear.

anarute:
Switching projects is not working very well (I'm using Chromium):

[Screencast from 2023-12-18 16-43-42.webm](https://github.com/Igalia/phpreport/assets/333447/16b9f10a-bd64-4b2e-a50c-0c8ce786acc3)

anarute:
> Switching projects is not working very well (I'm using Chromium):
> 
>  Screencast.from.2023-12-18.16-43-42.webm

This is working now

anarute (approved):
LGTM thanks!

PR #745, opened 2023-12-18
Allow templates with empty task types

anarute:
If a user had a template with no task type they couldn't use it.


dmtrek14 (approved):
LGTM

PR #748, opened 2023-12-18
Fix holidays calculations

dmtrek14:
- Change how we get used/scheduled hours if reference date is in the future
- Update calculation in summary report action to subtract used hours
- Rename variable to match the ui for clarity


anarute (approved):
LGTM

PR #749, opened 2023-12-18
Use http_build_query to build for the auth params

anarute:
In order to parse special characters and properly build the query params for the auth http request.


dmtrek14 (approved):
LGTM - thanks!

PR #750, opened 2023-12-18
Improve day view component

negreirosleo:
This PR focuses on improving the organization of the files for the tasks page so it's easier to understand what is going on, these are the main points:

1. Transform `TaskForm.tsx` into `DayView.tsx`, this is more inline of what that part of the page represents, the task listing / creation on a single day view, thus making the `TaskList` separate from the form. 

2. Break `useTask.ts` into different files, one for each hook. there was no `useTask` hook so it doesn't make any sense to have that file.


anarute (approved):
LGTM thanks!

dmtrek14 (approved):
Looks good - thank you!

PR #751, opened 2023-12-18
Adjust window of events to calendar  sync

dmtrek14:
- In order to avoid very large request sets to the calendar provider, the window here is reduced to 3 months in the past until Dec 31 of the next year.

### DIFF
--- a/model/facade/action/SyncCalendarAction.php
+++ b/model/facade/action/SyncCalendarAction.php
@@ -57,11 +57,11 @@ class SyncCalendarAction extends Action
 
         $client->setCalendar($arrayOfCalendars[ConfigurationParametersManager::getParameter('CALENDAR_ID')]);
 
-        $lastYear = date("Y") - 1;
         $nextYear = date("Y") + 1;
-        $start = gmdate("Ymd\THis\Z", strtotime($lastYear . "-01-01"));
-        $end = gmdate("Ymd\THis\Z", strtotime($nextYear . "-12-31"));
-        $currentEvents = $client->getEvents($start, $end);
+
+        $startSyncPeriod = gmdate("Ymd\THis\Z", date(strtotime("-3 months")));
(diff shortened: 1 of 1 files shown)


model/facade/action/SyncCalendarAction.php
```diff
@@ -57,9 +57,9 @@ protected function doExecute()
 
         $client->setCalendar($arrayOfCalendars[ConfigurationParametersManager::getParameter('CALENDAR_ID')]);
 
-        $lastYear = date("Y") - 1;
         $nextYear = date("Y") + 1;
-        $start = gmdate("Ymd\THis\Z", strtotime($lastYear . "-01-01"));
+
+        $start = gmdate("Ymd\THis\Z", date(strtotime("-3 months")));
         $end = gmdate("Ymd\THis\Z", strtotime($nextYear . "-12-31"));
         $currentEvents = $client->getEvents($start, $end);
```
anarute, line 64: I think it's better to rename these variables now that we are using the start inside the loop to avoid it being overwritten in the line 73.

```suggestion
        $startSyncPeriod = gmdate("Ymd\THis\Z", date(strtotime("-3 months")));
        $endSyncPeriod = gmdate("Ymd\THis\Z", strtotime($nextYear . "-12-31"));
        $currentEvents = $client->getEvents($startSyncPeriod, $endSyncPeriod);
```

model/facade/action/SyncCalendarAction.php
```diff
@@ -69,8 +69,7 @@ protected function doExecute()
         }
 
         foreach ($this->datesRanges as $range) {
-            if (!is_array($range) || !isset($range['start'])) continue;
-
+            if (!is_array($range) || !isset($range['start']) || $range['start'] < $start) continue;
```
anarute, line 72:
```suggestion
            if (!is_array($range) || !isset($range['start']) || $range['start'] < $startSyncPeriod) continue;
```

anarute (approved):
Thank you for the changes!

PR #753, opened 2023-12-20
Select template on task creation

negreirosleo:
Selecting a template will auto-fill the create task form:

[Screencast from 2023-12-20 09-14-44.webm](https://github.com/Igalia/phpreport/assets/20492786/2511755f-8ca8-4c6b-b7fe-17dad28be18d)


anarute (approved):
LGTM

frontend/src/domain/Task.ts
```diff
@@ -39,7 +39,7 @@ export const TaskIntent = z
   .object({
     userId: z.number(),
     projectId: z.string().min(1, { message: 'Project is required' }),
-    taskType: z.string().optional(),
+    taskType: z.string().nullable(),
     story: z.string(),
```
dmtrek14, line 43: The story and description will also be nullable. I am mentioning in case it is something you need to account for.

frontend/src/domain/Template.ts
```diff
@@ -1,9 +1,9 @@
...
   id: number
   name: string
-  story: string
-  description: string
-  taskType: string
+  story: string | null
+  description: string | null
+  taskType: string | null
   userId: number
   projectId: number
```
dmtrek14, line 8: The projectId and userId are also nullable in a template. When we start using global templates in the list, the userId will be null for those. A user's personal templates will have their userId.

frontend/src/app/tasks/hooks/useCreateTaskForm.ts
```diff
@@ -54,19 +56,31 @@ export const useTaskForm = () => {
...
+    const template = templates.find((t) => t.id === templateId)
+    if (template) {
+      setTemplateName(template.name)
+      setFormState((prevState) => ({
+        ...prevState,
+        taskType: template.taskType,
+        description: template.description || '',
+        startTime: template.startTime,
+        endTime: template.endTime,
+        projectId: template.projectId.toString(),
```
dmtrek14, line 69: I noted it elsewhere in the review, but the projectId in a template is nullable, so this will throw an error in the case it is null. I also wondered why we are using toString() on the projectId?

frontend/src/domain/Template.ts
dmtrek14, line 1: We are missing the `startTime` and `endTime` fields in the type here. These will be nullable here.

frontend/src/app/tasks/hooks/useCreateTaskForm.ts
negreirosleo: hmmm, it should be a number actually, I will fix it

dmtrek14 (approved):
Thanks for the changes - LGTM!

PR #754, opened 2023-12-20
Add button to expand task

negreirosleo:
After clicking the expand icon it shows description and story.

[Screencast from 2023-12-20 15-11-23.webm](https://github.com/Igalia/phpreport/assets/20492786/b599dc22-05ad-43cd-bf24-ecad67c5e1fa)

### DIFF
--- a/frontend/src/app/tasks/components/TaskBox.tsx
+++ b/frontend/src/app/tasks/components/TaskBox.tsx
@@ -4,15 +4,20 @@ import { useState } from 'react'
 import Box from '@mui/joy/Box'
 import Button from '@mui/joy/Button'
 import Typography from '@mui/joy/Typography'
-import { Delete24Filled, Edit24Filled } from '@fluentui/react-icons'
+import {
+  Delete24Filled,
+  Edit24Filled,
+  ArrowMaximize24Filled,
+  ArrowMinimize24Filled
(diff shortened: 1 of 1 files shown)


frontend/src/app/tasks/components/TaskBox.tsx
```diff
@@ -68,18 +74,39 @@ export const TaskBox = ({ task, projects, taskTypes, tasks }: TaskProps) => {
...
       </Box>
       <Box display="flex" gap="8px">
-        <IconButton onClick={() => setEditMode(true)}>
+        <IconButton
+          onClick={() => {
+            setEditMode(true)
+            setExpandedTask(false)
+          }}
+        >
           <Edit24Filled color="#2f3338" />
```
dmtrek14, line 93: At some point, I think we should extract all hard-coded hex values and use a stylesheet and variables (or some other centralized method) for colors/styles. If we were to ever want to change our icon colors, for instance, it would be easier to change the value once than across multiple files. I won't hold the PR on this, but it is something we probably want to consider doing soon before we add too much more code. The theme (colors, fonts, border radius, etc.) that we are using in the app will be part of what we want to use with other applications, so we will want to implement it in a way that can be shared.

dmtrek14 (approved):

anarute (approved):

PR #756, opened 2023-12-29
Fix template select not selecting the project

negreirosleo:
The select template function wasn't setting the projectName so it left the project select empty when prefilling the form with a template.


dmtrek14 (approved):
LGTM - thanks!

anarute (approved):

PR #757, opened 2023-12-29
Create template modal

negreirosleo:
Create save task as template modal.

Instead of having controlled inputs and react query I tried using the new server actions from next js v14 to submit the form. I like the result of it, it save us from writing more code to track the fetch states, and having to use more hooks.

[Screencast from 2023-12-29 18-23-09.webm](https://github.com/Igalia/phpreport/assets/20492786/039547b6-8327-4eb4-8152-e7eac06136ff)


frontend/src/infra/template/createTemplate.ts
```diff
@@ -0,0 +1,34 @@
...
+import { ApiClient } from '../lib/apiClient'
+import { BaseError } from '@/_lib/errors/BaseError'
+
+const createTemplateError = (message = 'Failed to create template') => {
+  const name = 'CreateTemplateError'
+
+  return new BaseError({ message, name, code: name })
+}
+
+export const makeCreateTemplate = (apiClient: ApiClient) => async (template: TemplateIntent) => {
```
anarute, line 11: It's a bit confusing that this is called `makeCreateTemplate` instead of `createTemplate` like we have the other infra methods: `createTask`, `deleteTask`, etc.

I would rename this one to `createTemplate` to keep the same pattern and rename the action `createTemplate` to `createTemplateAction`.

anarute (approved):
Other than the naming nit, looks good to me! I also tested locally and it works well.

dmtrek14 (approved):
I agree with @anarute on the naming, but this otherwise looks good to me.

frontend/src/infra/template/createTemplate.ts
negreirosleo: I named it `makeCreateTemplate` because its a function that creates another function. I actually started with `createTemplateAction` but I thought it could be mistaken by "creating a template-action".

This pattern for naming is common for function factories in the js environment, you can even check an example on this DI container lib:

https://github.com/jeffijoe/awilix

```ts
// Let's try with a factory function.
const makeUserService = ({ db }) => {
  // Notice how we can use destructuring
  // to access dependencies
  return {
    getUser: id => {
      return db.query(`select * from users where id=${id}`)
    }
  }
}
```

negreirosleo: Check how it looks on the code:
```ts
  const apiClient = await serverFetch()
  const createTemplate = makeCreateTemplate(apiClient)
  ```
  
  the `createTemplate` function is the function that is returned by `makeCreateTemplate`

anarute: Ok let's leave it like this then

PR #769, opened 2024-01-04
Add allocation model and endpoints

anarute:
The creation is a simple method but to get the allocation ready for the fronted, they are grouped by user and weeks instead of being returned raw as the DB model.

This is to avoid too many calculations in the frontend to display the data in a timeline.

In this phase of the project there are still too many business rules missing like what to do with overlapping allocations and overbooking, but this will be done in a later stage.

We also still need to add the permissions logic to allow only Project Managers to manage allocations.


api/schemas/project.py
```diff
@@ -18,4 +18,41 @@ class Project(BaseModel):
...
+    total_hours: Optional[float] = None
+    fte: Optional[float] = None
+    is_tentative: Optional[bool] = None
+    is_billable: Optional[bool] = None
+    notes: Optional[str] = None
+
     model_config = ConfigDict(alias_generator=to_camel, populate_by_name=True)
+
+
+class NewProjectAllocation(BaseModel):
```
dmtrek14, line 42: If all the properties are the same as the ProjectAllocation model, I think you can inherit. So `NewProjectAllocaiton(ProjectAllocation)` and then `pass` in the body. You would then only need to specify a property in the New model if it differed from the ProjectAllocation (i.e., if it was now required instead of optional).

dmtrek14 (approved):
Great start here. I only had one small suggestion, but it is more of a code style thought.

api/schemas/project.py
anarute: Good catch, they hadn't exactly the same properties but could be cleaned up indeed.

api/routers/v1/projects.py
```diff
@@ -28,3 +30,35 @@ async def get_projects(db: Session = Depends(get_db), offset: int = 0, limit: in
...
 async def get_project(project_id: int, db: Session = Depends(get_db)):
     return db.query(Project).filter(Project.id == project_id).first()
+
+
+@router.get("/{project_id}/allocations", dependencies=[Depends(BearerToken())])
+async def get_project_allocations(
+    project_id: int,
+    start: date = date.today(),
+    end: date = date.today(),
+    current_user=Depends(get_current_user),
```
dmtrek14, line 40: Do we need the `current_user` here? It doesn't appear that we are using any of its properties in the service call.

And will we worry about adding permissions a bit later? Totally fine if yes - just want to make sure it wasn't missed.

anarute: True, for now we are not using but we definitely will need for the permissions - only PMs should be able to manage the allocations

dmtrek14: Sounds good. We can add later as we continue to implement and refine.

PR #770, opened 2024-01-05
Add migration for config table

dmtrek14:
This PR adds a migration for the config table which will be later utilized for other work on the work summary panel. And because we will be starting to heavily use this table, additional constraints are added, such as only 1 record is allowed and deletes are not allowed to prevent accidental removal of config data.


anarute (approved):
LGTM, thank you!

PR #771, opened 2024-01-05
db: Add column to journey history table

dmtrek14:
In order to calculate and store the total amounts of vacation and hours expected in a year, we need a column in which to store that information. This migration adds the necessary column.


api/models/user.py
```diff
@@ -69,6 +70,7 @@ class UserCapacity(Base):
     start = Column("init_date", Date, nullable=False)
     end = Column("end_date", Date, nullable=True)
     user_id = Column("usrid", Integer, ForeignKey("usr.id"), nullable=False)
+    aggregates = Column("aggregates", JSONB, nullable=True)
```
anarute, line 73: Can we call it something else? Aggregates seems too generic and I'm having a hard time to understand what it will store

dmtrek14: Of course. It will hold the amount of vacation hours the person will have as well as the amount of total expected hours for each year in the time period. 
Screenshot from 2024-01-08 07-47-33
I had a hard time naming it (obviously ;)). If you don't mind being verbose, we could do `yearly_expected_and_vacation`or just `expected_and_vacation`. Or I am completely open to suggestions.

anarute: I don't mind verbose in this case, `yearly_expected_and_vacation` works for me, thank you!

anarute (approved):
LGTM, thanks!

PR #772, opened 2024-01-08
Clone task

negreirosleo:
This PR's adds a button to the TaskList that when clicked it clones the task data to the create task form

### Screenshots

[Screencast from 2024-01-08 11-42-49.webm](https://github.com/Igalia/phpreport/assets/20492786/8d42bb0a-1c7e-40f4-94db-cfd0d5e07cd7)


dmtrek14 (commented):
For cloning, the times are not copied in the existing application. The user will have to set them anyway or else it would fail validation as overlapping, so it is probably best to clear them here, too, to save the user a step.

I am also still noticing the issue where the tasks are not sorted correctly on the right side after I add a task.
Screenshot from 2024-01-09 08-51-24

I have also noticed that once a task is submitted, the form does not clear. Should we clear the form on successful submission? Now that we have clone  (which fills the form), it may start to get confusing to users if the form is essentially always filled. For instance, I started the app to test and created a task for today so that I could clone it. Because the form didn't clear after I saved the task, it was hard to tell if clone was initially doing anything. What do you think?

negreirosleo:
@dmtrek14 I agree with you on the form clear and the dates, it creates a better UX

negreirosleo:
@dmtrek14 I also took the opportunity and added a loading state to the submit button :+1:

anarute (approved):
LGTM, thank you!

PR #774, opened 2024-01-09
api: Add new property to schema

dmtrek14:
- Add missing user_id property to schema
- Fix capacity type
- Rename property in dependencies


anarute (approved):
LGTM

PR #775, opened 2024-01-10
db: Add utility functions

dmtrek14:
- Add functions to be used in calculation of yearly_expected_and_vacation for work summary

Each function is commented with purpose and any "gotchas" (such as needing to add a day in a date diff because one end isn't included). I decided to put the functions in their own folder because we are starting to get enough db infra to justify keeping track of these better. Eventually, I will also move the archiving function into this folder, as well. Having functions in their own files will allow us to more easily keep track of changes to them in source control.


api/db/functions/get_vacation_for_period.py
```diff
@@ -0,0 +1,48 @@
...
+            declare company_fte_weekly int := (select config.company_fte from public.config limit 1);
+            BEGIN
+                return(
+                    select (end_date - start_date + 1) /
+                        (
+                        make_date(date_part('year', end_date::date)::int, 12, 31) -
+                        make_date(date_part('year', start_date::date)::int, 01, 01) + 1
+                        )::float
+                * yearly_vacation_hours
+                * (capacity * 5 / company_fte_weekly::float)
```
anarute, line 41: Nit: I think the indentation is a bit off for these lines, it should be one tab more?

anarute (approved):
LGTM!

api/db/functions/get_vacation_for_period.py
dmtrek14: Good catch - fixing!

PR #780, opened 2024-01-17
db: Add migration for calculated field

dmtrek14:
Add calculated field to task table to do the end - init calculation at db level. This will be used by work summary calculations.


anarute (approved):
LGTM, thanks!

PR #781, opened 2024-01-19
Create week view

negreirosleo:
This PR creates the week view

### Screenshots

image

### DIFF
--- a/frontend/src/app/tasks/components/SaveTemplateModal.tsx
+++ b/frontend/src/app/tasks/components/SaveTemplateModal.tsx
@@ -1,64 +1,73 @@
+import { useState } from 'react'
 import { Sheet, Typography, Modal, ModalClose } from '@mui/joy'
-
 import { Input } from '@/ui/Input/Input'
 import { Task } from '@/domain/Task'
 import { createTemplate } from '../actions/createTemplate'
 import { useAlert } from '@/ui/Alert/useAlert'
 import { SubmitButton } from '@/ui/SubmitButton/SubmitButton'
+import { SaveCopy24Filled } from '@fluentui/react-icons'
+import { IconButton } from './IconButton'
 
 type SaveAsTemplateProps = {
-  closeModal: () => void
-  open: boolean
   task: Task
 }
 
-export const SaveTemplateModal = ({ closeModal, open, task }: SaveAsTemplateProps) => {
+export const SaveTemplateModal = ({ task }: SaveAsTemplateProps) => {
   const { showError, showSuccess } = useAlert()
+  const [templateModalOpen, setTemplateModalOpen] = useState(false)
+
+  const closeModal = () => setTemplateModalOpen(false)
 
   return (
-    <Modal
-      sx={{ display: 'flex', justifyContent: 'center', alignItems: 'center' }}
-      open={open}
-      onClose={() => closeModal()}
-      aria-labelledby="confirmation-modal-title"
-      aria-describedby="modal-content"
-    >
-      <Sheet
-        variant="outlined"
-        component="form"
-        action={async (formData: FormData) => {
-          const template = await createTemplate(task, formData)
-          if (template.error) {
-            showError(template.error)
-          } else {
-            showSuccess('Template created')
-          }
-          closeModal()
-        }}
-        sx={{
-          minWidth: 300,
-          minHeight: 200,
-          borderRadius: 'md',
-          padding: 3,
-          boxShadow: 'lg',
-          display: 'flex',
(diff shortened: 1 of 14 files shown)


frontend/src/app/tasks/week/page.tsx
```diff
@@ -1,3 +1,87 @@
...
+    }
+
+    return { ...acc, [task.date]: { tasks: [task], time: timeDiff } }
+  }, {})
+}
+
 export default async function WeekView() {
-  return <div>Week!</div>
+  const today = new Date()
+  const sevenDaysAgo = sub(today, { days: 6 })
```
dmtrek14, line 37: I am pretty sure that we would want to display the current ISO week instead of today - 6 days. This would ensure that the grid here is always consistent - i.e., always Mon - Sun. @anarute what do you think?

frontend/src/app/tasks/components/SaveTemplateModal.tsx
```diff
@@ -1,64 +1,74 @@
...
+        onClick={closeModal}
+        icon={SaveCopy24Filled}
+        description="save task as template"
+      />
+      <Modal
+        sx={{ display: 'flex', justifyContent: 'center', alignItems: 'center' }}
+        open={templateModalOpen}
+        onClose={closeModal}
+        aria-labelledby="confirmation-modal-title"
+        aria-describedby="modal-content"
```
dmtrek14, line 33: There is no element with the id of `modal-content` so this will toss an a11y  error.

dmtrek14 (commented):
I think there is a regression somewhere. The 'clone', 'expand', and 'save as template buttons are no longer working for me in the Day view.

The week view is not very mobile friendly at the moment. While the functionality is more of a priority than mobile friendliness, I am noting it so that you can keep it in mind as you refine in case there are some easy changes you can make as you go that would help in this regard. For instance, I cannot even scroll to the right to see all of the tasks for the week.
Screenshot from 2024-01-19 10-07-02

Ultimately, a layout more like below might work for mobile (but, again, functionality is a higher priority than mobile):
Screenshot from 2024-01-19 10-20-20

negreirosleo:
@dmtrek14 I fixed the save template modal, but I couldn't replicate errors with the other buttons, can you show a print or video of it happening?

negreirosleo:
@dmtrek14 I added the mobile view, it has a slider for each row if they have a lot of tasks:

[Screencast from 2024-01-19 16-32-38.webm](https://github.com/Igalia/phpreport/assets/20492786/2edab5f6-4683-4097-9f80-077db9cf2124)

I just don't know about the height difference when a task has a task type, what do you think about it?

dmtrek14:
@negreirosleo Okay, so the button malfunction looks to have been some local issue with my environment. I deleted all my images and rebuilt and things work okay now, so ignore the earlier comment.

Regarding the height difference for task types: I wonder if we should hide the task type on mobile, but I think we might still run into some styling inconsistencies if we have some projects with short names and other with long names. I am happy to also leave this particular problem for another day and we can think on it as you continue implementation.

negreirosleo:
I think hiding the task type for mobile is cool and regarding very long project names we could use a `text-overflow: ellipsis` and it would be: "proejct with very long name" -> "project with very lon...."

dmtrek14:
I think that would work and wouldn't negatively impact usability.

negreirosleo:
Screenshots of the ellipsis:
- Mobile
image

- Desktop
image

frontend/src/app/tasks/week/page.tsx
anarute: Yes, agree.

negreirosleo: Cool, I'm working on it!

anarute:
One thing I noticed is that if you create a task in the daily view and return to the week it is not updated, we need to refresh the page manually. It's fine if you want to solve this in a new pull request, but I thought it was important to mention.

frontend/src/app/tasks/week/page.tsx
negreirosleo: image


Like this?

frontend/src/app/tasks/components/SaveTemplateModal.tsx
negreirosleo: I removed that because I found no use for it :thinking: there's no need to describe the modal since we already associate it to the title

dmtrek14 (approved):
LGTM - thanks for the adjustments

anarute (approved):

PR #782, opened 2024-01-19
Enable leave projects to everyone

anarute:
Instead of having to manually assign everyone to all the leave projects, return them when fetching user's projects.
Also check if they can log time to a project before creating a task.


dmtrek14 (approved):
Works for me on this end. Thanks for the patch.

PR #783, opened 2024-01-22
db: Add function to get list of project work for summary

dmtrek14:
This function breaks out work by project daily and weekly for a user. This will be used to fetch the breakdown for display in the work summary panel.


anarute (approved):
LGTM!

PR #784, opened 2024-01-23
db: Add fxn/trigger to calculate vacation/expected

dmtrek14:
This function will run against each capacity/journey_history when one is inserted or updated. The migration is set up to make a change to all records in the table so that the calculation will run for existing data. The function will use the dates in the range and the journey, along with company fte and yearly vacation hours, to determine how much vacation a user should have in each year in the journey and how many expected hours a user should work in the year. 

The resulting data inserted into `journey_history.yearly_expected_and_vacation` will look like this: 
```json
[{"2023": {"expectedHours": 1456.0000, "availableVacation": 128.8}}, {"2024": {"expectedHours": 1467.2000, "availableVacation": 128.8}}]
```


anarute (approved):
LGTM!

PR #785, opened 2024-01-24
Create month view

negreirosleo:
This PR adds the Month view to /tasks

### Screenshots
Desktop:
image

Mobile:
image

### DIFF
--- a/frontend/src/app/tasks/month/page.tsx
+++ b/frontend/src/app/tasks/month/page.tsx
@@ -1,3 +1,154 @@
-export default function MonthView() {
-  return <div>Month!</div>
+import {
+  startOfMonth,
+  startOfWeek,
+  lastDayOfMonth,
+  lastDayOfWeek,
+  format,
+  eachDayOfInterval,
+  isSameDay,
+  isSameMonth
+} from 'date-fns'
+import { Box, Typography } from '@mui/joy'
+import { getTasksGroupedByDate } from '../actions/getTasksGroupedByDate'
+import { convertMinutesToTime } from '../utils/time'
+import { SimpleTaskBox } from '../components/TaskBox'
+import { Fragment } from 'react'
+
+const WEEK_DAYS = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday']
+
+export default async function MonthView() {
+  const today = new Date()
+  const firstDayOfFirstWeek = startOfWeek(startOfMonth(today))
+  const lastDayOfLastWeek = lastDayOfWeek(lastDayOfMonth(today))
+  const dateRange = eachDayOfInterval({ start: firstDayOfFirstWeek, end: lastDayOfLastWeek })
+
+  const groupedTasks = await getTasksGroupedByDate(
+    format(firstDayOfFirstWeek, 'yyyy-MM-dd'),
+    format(lastDayOfLastWeek, 'yyyy-MM-dd'),
+    500
+  )
+
+  return (
+    <>
+      <Typography sx={{ fontSize: { xs: 'lg', sm: 'xl4' } }} level="h1" textAlign="center">
+        {format(today, 'MMMM yyy')}
+      </Typography>
+      <Box
+        sx={{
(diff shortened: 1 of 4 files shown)


frontend/src/app/tasks/actions/getTasksWithDate.ts
```diff
@@ -0,0 +1,32 @@
+import { makeGetTasks } from '@/infra/task/getTasks'
+import { serverFetch } from '@/infra/lib/serverFetch'
+import { authOptions } from '@/app/api/auth/[...nextauth]/route'
+import { getServerSession } from 'next-auth'
+
+import { Task } from '@/domain/Task'
+
+import { convertTimeToMinutes } from '../utils/time'
+
+export const getTasksWithDate = async (startTime: string, endTime: string, limit?: number) => {
```
anarute, line 10: I'm a little confused about the naming, make it's better to call it `getTasksGroupedByDate` or something like that?

frontend/src/app/tasks/month/page.tsx
```diff
@@ -1,3 +1,134 @@
...
+              ':nth-of-type(7n)': { borderRight: { sm: 'none' }, borderLeft: { sm: 'none' } },
+              fontWeight: '600',
+              color: isToday ? '#004c92' : 'auto',
+              fontSize: { xs: '0px', sm: '1rem' },
+              ':first-letter': { fontSize: '1rem' }
+            }}
+            component="div"
+            key="day"
+          >
+            {isToday ? 'Today' : day}
```
dmtrek14, line 60: I am not sure we want the whole day of the week column to show as 'Today' instead of 'Wednesday'. I know the mockup shows this, but I think it is enough to have the slightly darker background for the day that is today. 
Screenshot from 2024-01-24 10-59-44

frontend/src/app/tasks/month/page.tsx
```diff
@@ -1,3 +1,134 @@
...
+  const dateRange = eachDayOfInterval({ start: firstDayOfFirstWeek, end: lastDayOfLastWeek })
+
+  const groupedTasks = await getTasksWithDate(
+    format(firstDayOfFirstWeek, 'yyyy-MM-dd'),
+    format(lastDayOfLastWeek, 'yyyy-MM-dd'),
+    500
+  )
+
+  return (
+    <Box
```
dmtrek14, line 32: Right above the grid holding the month, what do you think about adding an `<h1>` centered above it with the name of the month? It can be good to make sure users know where/when they are. And once we add the date picker at the top, I imagine it would be more important to have the month as a wayfinding helper.  I don't have a strong preference on full month name  or three-letter abbreviation. 

It would probably be nice on the week page, too: 
- January 21, 2023 - January 27, 2023  or 
- Jan 21, 2023 - Jan 27, 2023  or
- Jan 21, 2023 - Jan 27, 2023 (W4)

Having headings is also a good way to help users who rely on assistive devices to know what is on a page. Right now, the page title of the app only ever reads 'PhpReport' which is not terribly informative, so having a heading level 1 with the page title would be useful.

Screenshot from 2024-01-24 11-28-51

dmtrek14 (commented):
The only other thing I noticed is that the dates and hour totals for each day are a little squished on mobile sometimes. 

What do you think about the following for smaller devices:
- Drop the ordinal (st, nd, rd) from the day numbers
- Decrease the font size just a little for the hour totals (I think setting `level="title-sm"` on that `<Typography>` element would do it.
- Maybe exclude days from other months? Or else use 12/31 instead of Dec. 31st?

Screenshot from 2024-01-24 11-36-46

Here's what it might look like with the above changes:
Screenshot from 2024-01-24 11-49-28

negreirosleo:
@dmtrek14 How about this:

image


I don't think its necessary to remove the past and next month days

frontend/src/app/tasks/month/page.tsx
negreirosleo: Week:

image

Month:

image

dmtrek14 (approved):
LGTM - thanks for the changes!

PR #786, opened 2024-01-24
Add work summary panel

dmtrek14:
The work summary panel will be visible only from the tasks area and show the user some basic information about the time they have worked for today and the week, as well as vacation time and extra hours. This information is shown in the current version of the application on the left side of the app. Some small improvements have been made here, such as breaking down the worked hours by project for a user and adding tooltips to explain each vacation category (used, scheduled, pending). 

A few small UI changes have been made here, as well. The collapse button for the right panel has been modified to be within the panel instead of a floating button. The dark/light mode toggle has been made smaller on the desktop view (same as it appears on mobile).

This is a first version of the component and it might be nice to display the time per project from the 'Worked Hours' panel as a pie chart instead of (or in addition to?) the table. Another planned feature for the 'Vacation' panel would be to add a calendar that shows upcoming vacation at a glance. 

[Screencast from 2024-01-24 10-34-25.webm](https://github.com/Igalia/phpreport/assets/3666105/24a7223c-c33e-4757-9215-0fc324f0fbc3)

### DIFF
--- a/api/services/timelog.py
+++ b/api/services/timelog.py
@@ -1,11 +1,12 @@
 from typing import List
-from datetime import date, datetime
-from sqlalchemy import or_
+from datetime import date, datetime, timedelta
+from sqlalchemy import or_, func
 from fastapi.encoders import jsonable_encoder
 
 from services.main import AppService
 from models.timelog import TaskType, Template, Task
-from schemas.timelog import TemplateNew, TemplateUpdate, TaskNew, TaskUpdate
+from models.config import Config
+from schemas.timelog import TemplateNew, TemplateUpdate, TaskNew, TaskUpdate, ProjectTaskSummary
 from schemas.validation import ValidatedObject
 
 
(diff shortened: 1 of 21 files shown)


api/services/timelog.py
```diff
@@ -137,3 +139,73 @@ def check_task_for_overlap(self, task: Task) -> ValidatedObject:
...
+                today_text=total[3],
+                week_total=total[4],
+                week_text=total[5],
+                is_vacation=total[6],
+            )
+            if t.today_total is None:
+                t.today_total = 0
+            if t.week_total is None:
+                t.week_total = 0
+            project_totals_list.append(t)
```
anarute, line 174: You can simplify the None checks in the assignment directly and even skip creating a variable for this:

```suggestion
            project_totals_list.append(
                ProjectTaskSummary(
                    project_id=total[0],
                    project=total[1],
                    today_total=total[2] or 0,
                    today_text=total[3],
                    week_total=total[4] or 0,
                    week_text=total[5],
                    is_vacation=total[6],
                )
            )
```

api/services/timelog.py
```diff
@@ -137,3 +139,73 @@ def check_task_for_overlap(self, task: Task) -> ValidatedObject:
...
+        scheduled = (
+            self.db.query(func.sum(Task.task_total_minutes).label("vacation_sum"))
+            .filter(Task.project_id == config.vacation_project_id, Task.date.between(tomorrow, yearEnd))
+            .first()[0]
+        )
+        if scheduled is None:
+            scheduled = 0
+        return scheduled
+
+    def get_expected_worked_hours(self, user_capacities: List[UserCapacity], start_date: date, end_date: date) -> float:
```
anarute, line 202: This is a function that doesn't need anything from the parent Class. What do you think of moving it to the helpers so we can also use it in other places like the project planning feature?

api/services/timelog.py
```diff
@@ -137,3 +139,73 @@ def check_task_for_overlap(self, task: Task) -> ValidatedObject:
...
+        end_date = end_date + timedelta(days=1)
+        days_in_range = (start_date + timedelta(x) for x in range((end_date - start_date).days))
+        expected = 0
+        for day in days_in_range:
+            if day.weekday() < 5:
+                capacity = [x.capacity for x in user_capacities if day >= x.start and day <= x.end]
+                if capacity:
+                    expected = expected + capacity[0]
```
anarute, line 210: How about already getting the weekdays before instead of adding the condition to the second loop to avoid so much nesting in the for loop?
I personally like to keep a maximum of 2 nesting levels whenever is possible.

```suggestion
        weekdays_in_range = (start_date + timedelta(x) for x in range((end_date - start_date).days) if x.weekday() < 5)
        expected = 0
        for day in weekdays_in_range:
            capacity = [x.capacity for x in user_capacities if day >= x.start and day <= x.end]
            if capacity:
                expected = expected + capacity[0]
```

api/tests/routers/v1/test_timelog.py
```diff
@@ -574,3 +574,160 @@ def test_user_cannot_create_task_with_overlapping_hours(
...
+
+
+def test_get_summary_fulltime_worker_entire_year(
+    client: TestClient, get_regular_user_token_headers: Dict[str, str]
+) -> None:
+    # there are 260 work days in 2023, user_id 1 has a capacity of 8.0 hrs/day
+    # so 260 * 8 = 2080, which would be expected work hours for year
+    # the config entry in mock_data sets company alloted vacation to 200 hours
+    # The user has the 8.0 hr/day capacity for all of 2023, so they would receive
+    # the full 200 hours of vacation as alloted by the company
```
anarute, line 586: TIL alloted :smile:

api/services/timelog.py
dmtrek14: That's a great idea!

api/services/timelog.py
dmtrek14: Oh, that's so much better - thank you.

frontend/src/infra/workSummary/getWorkSummary.ts
```diff
@@ -0,0 +1,14 @@
...
+  apiClient: ApiClient
+): Promise<WorkSummary> => {
+  const response = await apiClient(`/v1/timelog/summary`)
+
+  if (!response.ok) {
+    throw new Error('Failed to fetch Work Summary')
+  }
+
+  return await response.json()
+}
```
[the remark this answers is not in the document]
dmtrek14: Yes! I will get that migrated. Thanks for catching that. I think I was working on the summary panel as you were doing the refactor and I forgot to go back and update.

frontend/src/ui/BaseLayout/BaseLayout.tsx
```diff
@@ -29,12 +30,15 @@ export const BaseLayout = ({ children }: { children: React.ReactNode }) => {
...
-      </ContentSidebar>
+      {
+        pathname.includes('/tasks') &&
+        <ContentSidebar
+          expanded={contentBarExpanded}
+          toggleContentBar={() => setContentBarExpanded((prevState) => !prevState)}
+        >
+            <WorkSummaryPanel contentSidebarExpanded={contentBarExpanded}/>
+        </ContentSidebar>
+      }
```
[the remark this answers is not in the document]
dmtrek14: We think we might use the panel for the project planner tool with some data that is similar to what is here. But I like the idea of moving this to a layout for the page rather than my shameless hack ;)

frontend/src/app/tasks/layout.tsx
```diff
@@ -1,14 +1,21 @@
...
+
 
 export default function TasksLayout({ children }: PropsWithChildren) {
+  const [contentBarExpanded, setContentBarExpanded] = useState(false)
+
   const pathname = usePathname()
 
   return (
+    <>
     <Box
```
negreirosleo, line 19: we should add a `margin: '0 60px 0 auto'` here in to not overlap the sidebar with the content:

with the margin:

image


without the margin:


image

negreirosleo (commented):
Besides the content overlap, it lgtm!

anarute (approved):

PR #789, opened 2024-01-30
Fix timer being unreliable

negreirosleo:
This unrealiablity was due to the setInterval behaviour, instead of adding one second each interval it compares the startTIme with the current time.

### DIFF
--- a/frontend/src/hooks/useTimer/useTimer.ts
+++ b/frontend/src/hooks/useTimer/useTimer.ts
@@ -1,22 +1,26 @@
-import { useState, useEffect, useCallback } from 'react'
+import { useState, useEffect, useCallback, useMemo } from 'react'
 
 export const useTimer = () => {
   const [time, setTime] = useState(0)
-  const [isTimerRunning, setIsTimerRunning] = useState(false)
+  const [startTime, setStartTime] = useState(0)
 
-  const startTimer = useCallback(() => setIsTimerRunning(true), [])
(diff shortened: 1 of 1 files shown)


frontend/src/hooks/useTimer/useTimer.ts
```diff
@@ -1,22 +1,26 @@
...
     setTime(0)
   }, [])
 
+  const isTimerRunning = useMemo(() => startTime > 0, [startTime])
+
   useEffect(() => {
     let intervalId: NodeJS.Timeout | undefined
     if (isTimerRunning) {
-      intervalId = setInterval(() => setTime((prevTime) => prevTime + 1), 1000)
+      intervalId = setInterval(() => setTime((Date.now() - startTime) / 1000), 1000)
```
negreirosleo, line 20: This seems like a small change,  but `setInterval` with 1000ms, it doesn't guarantee that the call back will happen everysecond, but it will happen in at least 1 second.

One exemple is this:

```ts
setInterval(() => console.log('a'), 1000)
while(true) {}
```

This console.log will never happen, because it is locked by the execution. So instead of adding 1 second by interval, it compares the current date with the start time

anarute (approved):
It looks good to me, thanks!

I left the timer running for almost my whole and it was accurate.

dmtrek14 (approved):
LGTM. I have been using the time most of the morning and it is working well for me.

PR #791, opened 2024-01-31
Sort leave projects last in tasks page dropdown

dmtrek14:
This PR adds a small sort to the projects list dropdown on the Tasks page. It sorts by type (so that projects of type 'leave' are sorted after those without a type) then by name.

### DIFF
--- a/web/js/tasks.js
+++ b/web/js/tasks.js
@@ -65,7 +65,8 @@ var projectRecord = new Ext.data.Record.create([
     {name:'id'},
     {name:'description'},
     {name:'fullDescription'},
-    {name:'customerName'}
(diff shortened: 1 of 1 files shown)


anarute:
I'm testing locally and it doesn't seem to be sorting by type correctly:

image

The last project in this list should be placed in third position as it doesn't have the type set as 'leave'. It does have the type set as 'per_hours', so I guess in the sorting you need to take into account the specific type 'leave' not just if it has a type set or not.

dmtrek14:
Okay, I tried doing it a little differently by assigning sort "points" based on type then description

anarute (approved):
LGTM, thanks!

PR #792, opened 2024-02-01
Add Date selector to tasks layout

negreirosleo:
This is the first step (of 2) in the date picker for /tasks. it adds the base for selecting dates, in the next PR the calendar will be added.

The date navigation works based on the url date param, eg: `/tasks/week/2024-01-24`, and this will show the of the 24th of January. 

If its a day view -> fetch all tasks for that day
If its a week view -> fetch all tasks for the date week
It its a month view -> fetch all tasks for the date month

Also, the page for the day view was changed from `/tasks` to `/tasks/day`, to follow the same pattern as month and view.

### Screenshots

[Screencast from 2024-02-19 14-40-35.webm](https://github.com/Igalia/phpreport/assets/20492786/3d19491f-9f47-4c3b-be1e-22394a62c0e2)


anarute:
Thank you for this!

In the mockup this was probably not very clear but the navigation should be per "page" not per day. So for example if you are in the Week view, when clicking the next arrow it should got to the next week directly not day by day. If you are in the Monthly view it should go to the next or previous month.

Check how google calendar works for example:

[Screencast from 2024-02-07 12-03-46.webm](https://github.com/Igalia/phpreport/assets/333447/e03615b0-1019-41dd-b09a-a0b0a0b10116)

anarute (approved):
LGTM! The date should be a date picker instead of a static date text, though I'm happy to merge this PR and have a new one adding the date picker.

dmtrek14 (approved):
In addition to what Ana already pointed out, the only thing I noticed is that there is sometimes a bit of a lag loading the data. For instance, I had tasks on Jan 2, 4, 9, and 10. If I used the arrows next to the date picker to click through dates, sometimes it would take several seconds for the tasks to load, but this behavior was not entirely consistent. I am wondering if a skeleton loader for the task cards would make sense here to indicate to the user that data is being loaded. This could be done in a future PR as a refinement. Thoughts?

anarute (approved):

dmtrek14 (approved):

PR #793, opened 2024-02-09
Expand user model and start getting user details from login token

anarute:
- Add 4 new fields to the user model: first name, last name, is_active and email.
- Start fetching these details from the user token and store them in the db
- Update user details in case name or email don't match


dmtrek14 (approved):
Looks good here!

PR #795, opened 2024-02-09
Add customer column to user detail report

dmtrek14:
This PR adds the Customer name as a column to the 'User Project Worked Hours' (aka UserDetails) report so that project managers can see the customer for the project.


anarute (approved):
LGTM, thanks!

PR #796, opened 2024-02-12
patch: Add envs for port and host, fix endpoint

dmtrek14:
This PR contains 2 small patches to fix errors found on deployment.

- Not specifying the port/hostname as env values in frontend image works when running locally but fails in a production environment
- Fix task type endpoint


anarute (approved):
LGTM

PR #804, opened 2024-03-08
Add total hours to reports

dmtrek14:
This PR adds total hours to two existing report: viewUserDetails and userTaskReport. 

Main changes include: 
- Add a grid footer where we will place the total
- On store load, append data total to the footer
- Style the footer
- Change the grid type in the userTasksReport from simple grid panel to ExportableGridPanel
- Update the ExportableGridPanel to export the total row to the csv if it is present in the grid

I forgot to turn on my 'format on save' in VSCodium, so it appears there are a lot of changes to userTasksReport. The actual code changes are at:
- Lines 423-428: add function to get minutes and variable to hold the total string
- Lines 442-454: add listener for load of store and append total hours to the footer
- Line 524: change the type of grid panel
- Line 550: add the style for the footer, remove the previous 'download as csv' button. 

For this last change, the button is removed because I changed the grid panel type to ExportableGridPanel which already has the button in the component.

### DIFF
--- a/web/js/include/ExportableGridPanel.js
+++ b/web/js/include/ExportableGridPanel.js
@@ -24,29 +24,33 @@
  * @param columnModel with column data.
  * @return string in CSV format.
  */
-function fromStoreToCSV(store, columnModel) {
-    var csv = "";
+function fromStoreToCSV(store, columnModel, totalRow = null) {
+  var csv = '';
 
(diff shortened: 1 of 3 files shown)


anarute (approved):
LGTM! I tested locally and it works fine, thanks!

PR #805, opened 2024-03-12
Add dates to holiday summary report header

dmtrek14:
This adds some additional information to the table header in the holiday summary report - specifically, it adds the first day of the week. This change is made so that users won't have to look up when an ISO week starts and can then have a better idea of when leaves are happening.

Once again, some of the changes show here are from auto-formatting - sorry! 

Main change in HolidayService.php is in lines 420-426 where the new array is created that has the week ('2024W01') and the weekStartDate ('Jan 1').

Main changes in holidaySummary.js are at lines 36 and 105 where the weeksStartDate is added and filled.

Then in the holidaySummary.php, the change is at line 69 where the additional info (the week start date) is added to the table header.

### DIFF
--- a/web/holidaySummary.php
+++ b/web/holidaySummary.php
@@ -66,7 +66,7 @@ include_once("include/header.php");
                 <th>Scheduled (hours)</th>
                 <th>Pending (hours)</th>
                 <th>% planned</th>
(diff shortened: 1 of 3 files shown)


anarute (approved):
LGTM

PR #809, opened 2024-05-07
Use scheduled hours instead of used for percentage

anarute:
Used hours are only for scheduled hours that are in the past. To calculate percentage it is needed to use all the scheduled hours for the period instead.


dmtrek14 (approved):
lgtm

PR #814, opened 2026-04-11
fix: redirect user to previous URL after auth (#779)

### LINKED ISSUE #779: Redirect user to previous url before auth
If a user is accessing a url directly or their session expires, after they sign in again they are redirect to `/tasks`. Ideally they should be redirected to the url they tried to access or where they were before the authentication.

TheSeydiCharyyev:
## Description

  When a user navigates to a specific page (e.g. `/planner`) and their session has expired (or they are not logged in), after signing in they are always redirected to `/tasks` instead of the page they originally wanted. This PR preserves the original URL through the auth flow.

  ## Related Issue

  Closes #779

  ## Changes Made

  - **`middleware.ts`** — added `callbackUrl` query parameter (set to the current path) when redirecting unauthenticated users to the signin page
  - **`authOptions.ts`** — updated the `redirect` callback to return `url` when it starts with `baseUrl`, instead of always returning `baseUrl`
  - **`signin/page.tsx`** — reads `callbackUrl` from search params and passes it to `signIn()`. Also fixes a typo: `keycloack` → `keycloak`

  ## How to Test

  1. Run `docker compose -f docker/docker-compose.dev.yml up`
  2. Open http://localhost:5173/planner (or any non-tasks page) without being logged in
  3. After signing in, verify you are redirected to `/planner` instead of `/tasks`


anarute:
Thanks for the interest but the project is no longer taking contributions.

